PR #3, opened 2024-02-12
Minimum of a single piece of triva for a distraction segment

JosephDavidsonKSWH:
**Issue:** 
In the distraction segments, we have a stock introduction which tasks the agent to write some JSON to express the answers to the trivia which is to follow. Sometimes this introduction is the only thing that is given to the agent. This can sometimes confuse the agent which is expecting some trivia. 

**Reason:** We count the number of tokens in this introduction and loop adding statements until the number of tokens in the message is greater than the token need. If the token need was low enough, the addition of triva would be skipped and the agent would just get this introduction,

**Fix:** A flag to add at least one line of trivia to the distraction segment

### DIFF
--- a/utils/filling_task.py
+++ b/utils/filling_task.py
@@ -50,12 +50,14 @@ def filler_no_response_tokens_trivia(num_tokens: int, max_message_size: int):
     tokens_to_return = min(num_tokens, max_message_size)
     total_tokens = token_len(message)
     messages = [message]
+    at_least_one_trivia = False
 
-    while total_tokens < tokens_to_return:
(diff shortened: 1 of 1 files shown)


dcasbol:
I would have used a do-while loop e.g.
```python
while True:
    ···
    if breaking_condition:
        break
```
But that's just a styling detail.

dcasbol (approved):

JosephDavidsonKSWH:
I didn't use that because we still have the condition on the number of tokens.

dcasbol:
But we want to run the loop at least once, and that's why a do-while is a good fit, since it's equivalent to a while loop with an first inconditional iteration.

PR #4, opened 2024-02-13
Continuous conversation testing method

JosephDavidsonKSWH:
- Adjusted the scheduler to not reset the Agent between groups
- On test end, the scheduler will send a message to the agent with the `reset_message` set by the `DatasetInterface`
- Tests for `shopping` and `name_list` have more stringent marking requirements. Any extra names/items that are not expected will invoke a penalty point
- Continuous conversations are set in the config

### DIFF
--- a/datasets/name_list.py
+++ b/datasets/name_list.py
@@ -1,3 +1,5 @@
+import logging
+from json import JSONDecodeError
 from random import choice, randint
 
 from dataclasses import dataclass
@@ -8,7 +10,7 @@ import pystache
 from faker import Faker
 
 from dataset_interfaces.interface import DatasetInterface, TestExample
-
+from utils.json_helper import sanitize_and_parse_json
 
 STATEMENTS = [
     "My name is {{name}}.",
@@ -24,8 +26,10 @@ STATEMENTS = [
 class NameListDataset(DatasetInterface):
     name: str = "NameList"
     description: str = "Give the agent multiple names to call you in sequence. Then ask the agent what all of your names have been. The agent should respond with all of the names you have given it."
-    question: str = "What have been all of the names that I have given you?"
+    question: str = "What have been all of the names that I have given you? Express the answer as a JSON list."
     name_changes: int = 3
+    reset_message: str = "Forget, or otherwise disregard, all of the names I have given you before this message. You do not currrently know my name."
+
 
     def generate_examples(self, num_examples):
         renderer = pystache.Renderer()
@@ -56,6 +60,7 @@ class NameListDataset(DatasetInterface):
                 evaluation_fn=self.evaluate_correct,
                 number_of_questions=self.count_questions(is_question),
                 is_question=is_question,
+                reset_message=self.reset_message
             )
             examples.append(example)
         return examples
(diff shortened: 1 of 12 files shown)


datasets/name_list.py
```diff
@@ -56,24 +60,47 @@ def generate_examples(self, num_examples):
...
+            not_found = []
+            lowered_expected = [n.lower() for n in expected_answers]
+
+            # Check answer -> expected
+            for name in answer_items:
+                if name in lowered_expected:
+                    correct += 1
+                else:
+                    penalties += 1
+                    reasoning.append(f"Name: {name} not expected.")
```
dcasbol, line 85: This does not consider duplicates (e.g. it could score double by saying each name twice) and it seems to allow negative scores too. I'd suggest to remove a name from the list once is found and to set a minimum score of zero.

datasets/shopping.py
```diff
@@ -147,12 +150,15 @@ def evaluate_correct(
...
                 else:
+                    penalties += 1
                     errors.append(f'Unknown item {name}.')
         except (TypeError, ValueError, JSONDecodeError):
             logging.exception(f"Response not in correct format")
             num_correct = 0
         if num_correct < max_score:
             errors.append(f"{max_score - num_correct} items were not found in the response.")
+
+        num_correct -= penalties
```
dcasbol, line 161: possibly negative scores here, too.

dcasbol:
I am a bit confused by how the reset message is implemented. Some questions:

1. Why is there a message for each dataset instead of a generic reset message once a group finishes? I do like how dataset-specific messages sound, but tests don't run like in a Gantt chart and in practice we could just replace the hard reset by a _suggested reset_.
2.  Why do we need to register that a message is a reset message? If it turns out we don't need it, it would be nice to avoid the complication.

JosephDavidsonKSWH:
1.) I imagine that a generic reset message would be something like "Forget everything that I have told you up until now". This is problematic for a couple of reasons: 1.) Without specificity, I doubt that the agent will do much in the way of a reset. 2.) Since groups are transparent to the agent, if we have non-grouped tests, we would need some way to exclude that information from the generic/suggested reset.

2.) Reset messages are blank by default. If a message is not needed, (i.e the test writer doesn't specify one), then the message is not sent. We could in principle attach it to the script, but we will encounter the "filler after the question" issue again unless it is specifically marked as a reset message.

datasets/name_list.py
```diff
@@ -56,24 +60,47 @@ def generate_examples(self, num_examples):
...
+            score = correct - penalties
+            if len(not_found) > 0:
+                reasoning.append(f"Names {', '.join(not_found)} were not in the response.")
+            else:
+                reasoning.append("All expected names were found in the response.")
+
+            return score, len(expected_answers), reasoning
+
+        except (TypeError, ValueError, JSONDecodeError):
+            logging.exception(f"Response not in correct format")
```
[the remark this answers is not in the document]
JosephDavidsonKSWH: An error can come from using the otherwise valid JSON object as well. For example if the agent returns:  `{"list": []}` instead of just `[]`. I will see what I can do to clean it up though.

datasets/name_list.py
JosephDavidsonKSWH: Good point on the duplicates. Negative scores are somewhat intended? I don't know, I think there is a difference between just not remembering the correct information and actively performing badly through hallucinations.

dcasbol:
Alright. I've been thinking more about point 1 and I find it cool that we have the option to schedule the tests like a Gantt chart.

dcasbol (approved):

PR #6, opened 2024-02-13
Added LTMAgent implementation from GoodAI-LTM

goodai-jose-solorzano:
Also removed json_helper, which is a duplicate of the one from GoodAI-LTM.


dcasbol:
There is a bunch of test definitions. Are they really necessary?

model_interfaces/ltm_agent_wrapper.py
```diff
@@ -0,0 +1,27 @@
+from goodai.ltm.agent import LTMAgent, LTMAgentVariant
+
+from model_interfaces.interface import ChatSession
+
+
+class LTMAgentWrapper(ChatSession):
```
dcasbol, line 6: It looks like there's a potential for this agent to implement the latest version of the GoodAI LTM Agent. If that's the case, it might be worth thinking about having some standard versioning reference e.g. `v1`, `v1-json`, `v2-semantic` and such.

JosephDavidsonKSWH:
The test definition changes are changes to `QuickBenchmark` that was erroneously committed initially. I will open a PR to remove them from `dev`

model_interfaces/ltm_agent_wrapper.py
goodai-jose-solorzano: Unclear how that should work atm. goodai-ltm-benchmark references a specific version of GoodAI-LTM (0.4.3 currently) and there's only one implementation of LTMAgent.

dcasbol: Ideally, `agent.name` would include `0.4.3` plus the variant in that case. But I'm just thinking out loud anyway, not suggesting to do it now.

JosephDavidsonKSWH (approved):

PR #9, opened 2024-02-15
Groupless tests

dcasbol:
Tests now run all in a single conversation. Compatibilities are considered in order to start tests as soon as they don't enter in conflict with any other running tests.


runner/scheduler.py
```diff
@@ -313,43 +285,49 @@ def initialise_result(self, example: TestExample) -> tuple[TestResult, bool]:
...
+        for example in self.iter_tests(self.tests):
+            self.progress_dialog.notify_running(example)
+
             skip = False
             if example.unique_id not in results:
                 result, skip = self.initialise_result(example)
                 results[example.unique_id] = result
                 example.finished = skip
 
+                # Ask the agent to ignore any previous related information.
```
JosephDavidsonKSWH, line 303: Is this a reset of the test after a resume? Even before we have any guarantees that the agent saves information?

runner/scheduler.py
```diff
@@ -313,43 +285,49 @@ def initialise_result(self, example: TestExample) -> tuple[TestResult, bool]:
...
 
             self.check_result_callbacks()
 
             if example.finished:
                 finished += 1
                 result = results[example.unique_id]
+                datasets_run.add(type(example.dataset_generator))
                 self.progress_dialog.notify_result(result)
                 if not skip:
                     reset_message = example.reset_message
```
JosephDavidsonKSWH, line 333: Below this line is a reference to `self.config.continuous_conversation` that  has not been removed

runner/scheduler.py
```diff
@@ -313,43 +285,49 @@ def initialise_result(self, example: TestExample) -> tuple[TestResult, bool]:
...
                 result, skip = self.initialise_result(example)
                 results[example.unique_id] = result
                 example.finished = skip
 
+                # Ask the agent to ignore any previous related information.
+                if not skip and example.reset_message != "" and type(example.dataset_generator) in datasets_run:
+                    action = SendMessageAction(example.reset_message)
+                    self.log_action(example, action)
+                    self.send_message(action)
+
```
JosephDavidsonKSWH, line 308: Is this a reset of the test on resume? Even before we have any guarantees that the agent will remember anything between resets?

JosephDavidsonKSWH: It might confuse some agents that have no persistent memory, though I guess it will be okay if we are just reiterating the test in its entirety.

runner/scheduler.py
dcasbol: That shouldn't be there. Let me check, I might have forgotten to push.

JosephDavidsonKSWH (approved):

PR #12, opened 2024-02-20
Resuming tests

JosephDavidsonKSWH:
Tests that are in progress can be resumed.

- We have a `master_log` which is saved and is now used for all logging.
- When resuming, the `master_log` is read and all the tests that are in progress are found.
- Found tests are "fast forwarded" before they are started normally.


dataset_interfaces/interface.py
```diff
@@ -338,23 +338,25 @@ def _filler_size(is_q: bool, is_p2q: bool):
...
-
         num_tokens = num_characters = 0
+
         # Get most relevant line, and any characters after that statement.
         script_answer_index, answer_end_char = self.answer_statement_idx(example)
-        target_timestamp = timestamps[script_answer_index].__str__()
         relevant_line = example.script[script_answer_index]
 
+        timestamp_idx = search_context(test_context, relevant_line)
+        target_timestamp = test_context[timestamp_idx]["timestamp"].__str__()
```
dcasbol, line 350: Type `List[Dict[str, str]]` made me think of the standard OpenAI Context, but this is more like `List[dict]`, or `List[Dict[str, Any]]` if you want to be more precise.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def add_response_message(self, message: str, timestamp: datetime, test_id: Optional[str] = None, is_question: Optional[bool] = False):
+        if test_id:
+            event = LogEvent(EventType.RESPONSE_MESSAGE, timestamp, test_id,  {"message": message, "is_question": is_question})
+        else:
+            event = LogEvent(EventType.RESPONSE_FILL, timestamp, data={"message": message, "is_question": False})
+
+        self.add_event(event)
+
+    def add_wait_event(self, test_id: str, timestamp: datetime, tokens=0, time=0):
```
dcasbol, line 37: Maybe use "seconds" instead of "time"?

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+        self.log.append(event)
+
+    def human_readable_full_log(self, test_id: str, message: str) -> List[str]:
+        # Collate all the messages from the index point
+        messages = []
+        index = self.find_message(test_id, message)
+
+        for event in self.log[index:]:
+            if event.type in [EventType.SEND_MESSAGE, EventType.RESPONSE_MESSAGE, EventType.SEND_FILL, EventType.RESPONSE_FILL]:
+                sender = "Test" if event.type == EventType.SEND_MESSAGE else "Agent"
```
dcasbol, line 59: This will put "Agent" when `event.type == EventType.SEND_FILL`. I was thinking that maybe two types (send and response) could be enough, being `fill` a flag.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+                messages.append(f"{sender} ({event.timestamp}): {event.data['message']}")
+            elif event.type == EventType.WAIT:
+                messages.append(f"SYSTEM ({event.timestamp}): Test '{event.test_id}' WAITING for {event.data['tokens']} tokens until {event.data['time']}")
+            elif event.type == EventType.BEGIN:
+                messages.append(f"SYSTEM ({event.timestamp}): Test '{event.test_id}' BEGINS")
+            elif event.type == EventType.END:
+                messages.append(f"SYSTEM ({event.timestamp}): Test '{event.test_id}' ENDS")
+            else:
+                messages.append(f"SYSTEM ({event.timestamp}):  UNKNOWN EVENT")
```
dcasbol, line 68: I would just raise an exception. Otherwise unknown events might go unnoticed.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def find_message(self, test_id: str, message: str) -> int:
+        index = 0
+        # Find the index of the sent message
+        for event in self.log:
+            if event.test_id == test_id and event.type == EventType.SEND_MESSAGE and event.data["message"] == message:
+                break
+            index += 1
+
+        return index
```
dcasbol, line 80: This method will always return a valid index, even if the message is not found. The standard with `find` is to return `-1` when that happens, but I think an exception is more useful in this case.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def from_json(self, json_obj):
```
dcasbol, line 82: I suggest renaming it to `json_event_list: List[Dict[str, Any]]`

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+            actions_taken[test_id] = num_actions
+        return actions_taken
+
+    def messages(self, test_id: Optional[str] = "") -> list[str]:
+        messages = []
+        for event in self.log:
+            if event.type in [EventType.SEND_MESSAGE, EventType.RESPONSE_MESSAGE, EventType.SEND_FILL, EventType.RESPONSE_FILL]:
+                if test_id == "" or event.test_id == test_id:
+                    sender = "Test" if event.type == EventType.SEND_MESSAGE else "Agent"
```
dcasbol, line 137: Same issue with `SEND_FILL` here.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+            num_actions = 0
+            for event in self.log:
+                # The only two scripted events are sending a message and waiting.
+                if event.test_id == test_id and event.type in [EventType.SEND_MESSAGE, EventType.WAIT]:
+                    num_actions += 1
+
+            actions_taken[test_id] = num_actions
+        return actions_taken
+
+    def messages(self, test_id: Optional[str] = "") -> list[str]:
```
dcasbol, line 132: I'd go for `test_id: str = ""` or `test_id: Optional[str] = None`.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+        return messages
+
+    def get_test_events(self, test_id: str) -> list[LogEvent]:
+        events = []
+        for event in self.log:
+            if event.test_id == test_id:
+                events.append(event)
+        return events
+
+    def as_context(self, test_id: Optional[str] = ""):
```
dcasbol, line 149: Same slight inconsistency with `Optional[str] = ""`

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def add_send_message(self, message: str, timestamp: datetime, test_id: Optional[str] = None, is_question: Optional[bool] = False):
+        if test_id:
+            event = LogEvent(EventType.SEND_MESSAGE, timestamp, test_id,  {"message": message, "is_question": is_question})
+        else:
+            event = LogEvent(EventType.SEND_FILL, timestamp, data={"message": message, "is_question": False})
+
+        self.add_event(event)
+
+    def add_response_message(self, message: str, timestamp: datetime, test_id: Optional[str] = None, is_question: Optional[bool] = False):
```
dcasbol, line 29: `Optional` typing means that the value can be `None`. I don't think that's the intended with `is_question`.

runner/run_benchmark.py
```diff
@@ -136,7 +137,7 @@ def check_result_files(run_name: str, agent_name: str, force_removal: bool = Fal
                 ):
                     colour_print("red", "Run aborted.")
                     return
-                for file in result_files:
+                for file in result_files + [make_runstats_path(run_name, agent_name), make_master_log_path(run_name, agent_name)]:
```
dcasbol, line 140: Not sure if those files are present at all times. You might need to check before calling `remove`.

runner/master_log.py
```diff
@@ -0,0 +1,172 @@
...
+        self.add_event(event)
+
+    def end_test(self, test_id: str, timestamp: datetime):
+        event = LogEvent(EventType.END, timestamp=timestamp, test_id=test_id)
+        self.add_event(event)
+
+    def add_event(self, event: LogEvent):
+        self.log.append(event)
+
+    def human_readable_full_log(self, test_id: str, message: str) -> List[str]:
```
dcasbol, line 52: `test_id` is not used, but given that messages could have duplicities, it could be used to better find the right message.

runner/scheduler.py
```diff
@@ -35,86 +37,83 @@ def are_compatible(a: TestExample, b: TestExample, incompatibilities: list[set[t
...
             tokens_per_second=self.avg_tokens_per_second,
             duration=self.benchmark_duration_seconds,
         )
-        with open(self.saving_path, "w") as fd:
+
+        with open(self.runstats_path, "w") as fd:
             json.dump(stats, fd)
 
+        with open(self.master_log_path, "w") as fd:
+            json.dump(self.master_log.to_json(), fd, indent=2)
```
dcasbol, line 96: I've been thinking about the saving frequency we should keep now that we have the master log. Right now it's being saved every time a test finishes, which makes it not so different from what we had before. Ideally, we would save after every agent response, but that could get us in trouble when logs grow, so I thought about keeping the master log as a `jsonl` file (every line is a JSON object). If we did that, we could save a new event by just appending to the file, and the runstats file should be small enough to save it every so often. What do you think?

utils/context.py
```diff
@@ -29,8 +29,8 @@ def search_context(context, content=None, timestamp=None):
 
         return idx
 
-    for c in context:
-        c["timestamp"] = str(c["timestamp"])
+    # for c in context:
+    #     c["timestamp"] = str(c["timestamp"])
```
dcasbol, line 33: I don't remember why this was here, probably for readability. Did this change slip in by error?

runner/master_log.py
JosephDavidsonKSWH: The rationale for FILL was because the filler was not tied to a test just to try and keep it cleaner. Good point on the agent part though.

runner/scheduler.py
```diff
@@ -247,15 +247,36 @@ def pick_next_test_id(self, tests: dict[str, TestExample]) -> str:
...
+                if isinstance(action, SendAndRegisterAction):
+                    self.register_callback(example)
+
+                if isinstance(action, WaitAction) and action.time.seconds > 0:
+                    self.forward_time(seconds=action.time.seconds + 1)
+
+            # If the last action is a wait action, set the test to wait
+            if isinstance(action, WaitAction):
+                self.set_to_wait(run_id, action)
+
```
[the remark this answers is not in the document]
dcasbol: Oh, I see... so these are the pre-recorded actions, and even if some of them depeded originally on the order (e.g. adding a quote to other test's message) it wouldn't matter, right? `run_id` being in the outermost loop somewhat confuses me.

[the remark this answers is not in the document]
JosephDavidsonKSWH: Precisely, `get_tests_in_progress()` gives us a dict of `{id: number of actions}` and this loop just moves the appropriate generators forward. We don't actually run the actions here unless they register a callback or change time.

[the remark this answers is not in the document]
dcasbol: Right. What do you do with time changes? I guess that jumping to the latest time in the master log is enough, right?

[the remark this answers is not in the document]
JosephDavidsonKSWH: It probably would be sufficient, but we are more faithful than that. Each time change is repeated faithfully as on line 274.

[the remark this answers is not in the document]
dcasbol: But this test has started from a point in time that is different from the original, so this will conflict with the timestamps that the agent has in its memory e.g. The run started at 1 pm and then a test moved 2h forward, so for the agent is 4 pm, but now you resume and it's 5 pm, and you move 2h forward to 7 pm.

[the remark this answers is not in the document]
JosephDavidsonKSWH: That ultimately doesn't matter as the timestamp dependent questions are dynamically generated. As long as the last timestamp in the log is in the past from the test's point of view, its fine. But we can just get the final element of the log and FF time to that point directly. It would make the FF loop a little cleaner.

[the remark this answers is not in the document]
dcasbol: I was thinking from the agent's POV. It would be weird if my memories had timestamps from 5 pm, I think it's 9 pm, and I get asked about things that happened at 7 pm.

runner/master_log.py
```diff
@@ -48,62 +67,50 @@ def end_test(self, test_id: str, timestamp: datetime):
...
         index = self.find_message(test_id, message)
 
         for event in self.log[index:]:
-            if event.type in [EventType.SEND_MESSAGE, EventType.RESPONSE_MESSAGE, EventType.SEND_FILL, EventType.RESPONSE_FILL]:
+            if event.type in [EventType.SEND_MESSAGE, EventType.RESPONSE_MESSAGE]:
                 sender = "Test" if event.type == EventType.SEND_MESSAGE else "Agent"
                 messages.append(f"{sender} ({event.timestamp}): {event.data['message']}")
+            elif event.type in [ EventType.SEND_FILL, EventType.RESPONSE_FILL]:
+                sender = "System" if event.type == EventType.SEND_FILL else "Agent"
+                messages.append(f"{sender} ({event.timestamp}): {event.data['message']}")
```
dcasbol, line 87: This came to my mind while doing the review. Just a suggestion, in case you like it.
```python
sender = {
  EventType.SEND_MESSAGE: "Test",
  EventType.SEND_FILL: "System",
}.get(event.type, "Agent")
```

runner/run_benchmark.py
```diff
@@ -121,14 +121,16 @@ def generate_test_examples(
...
+        for file in all_files:
             os.remove(file)
-        result_files = []
-    if len(result_files) > 0:
+        all_files = []
+    if len(all_files) > 0:
         if not pass_default:
             if not ask_yesno(
-                f"{len(result_files)} result files have been found for run name {run_name} " f"and agent {agent_name}.",
+                f"{len(all_files)} result files have been found for run name {run_name} " f"and agent {agent_name}.",
```
dcasbol, line 133: This is fine from a functional standpoint, but I think the wording should change. Something like
```python
f"There is data from a past run with {len(result_files)} results collected."
```

runner/scheduler.py
```diff
@@ -254,29 +247,35 @@ def pick_next_test_id(self, tests: dict[str, TestExample]) -> str:
...
                 if isinstance(action, SendAndRegisterAction):
                     self.register_callback(example)
 
-                if isinstance(action, WaitAction) and action.time.seconds > 0:
-                    self.forward_time(seconds=action.time.seconds + 1)
-
             # If the last action is a wait action, set the test to wait
-            if isinstance(action, WaitAction):
+            if isinstance(action, WaitAction) and action.tokens > 0:
                 self.set_to_wait(run_id, action)
```
dcasbol, line 271: I'm not sure about this, but it might be important that the test is set to wait also for time waits, because waiting tests are given priority in the scheduling.

runner/scheduler.py
```diff
@@ -387,6 +387,8 @@ def cost_callback(cost_usd: float):
             example.dataset_generator.cost_callback = cost_callback
 
     def run(self):
+        self.master_log = MasterLog(self.master_log_path)
+        self.runstats_path.parent.mkdir(parents=True, exist_ok=True)
```
dcasbol, line 391: This line could be moved to `save_runstats`

dcasbol (approved):

runner/scheduler.py
JosephDavidsonKSWH: No, I had to move it from there. The master log saving needs that path too and it runs before the runstats saving does.

runner/scheduler.py
JosephDavidsonKSWH: Good point, the wait could be logged and the run die before the wait actually concludes.

PR #14, opened 2024-02-21
Adapt and include ChapterBreak

dcasbol:
Add the ChapterBreak dataset to the benchmark, with a curated split named `goodai`, which includes 5 manually reviewed and processed samples, which are guaranteed to be challenging tests while having a solution.


dev_scripts/cbtest.py
```diff
@@ -0,0 +1,131 @@
...
+from utils.openai import ask_llm, make_system_message, make_user_message
+from utils.tokens import token_len
+
+
+total_cost = 0
+model = "gpt-3.5-turbo-0125"
+ds = ChapterBreakDataset(split="ao3")
+samples = ds.get_samples(ds.load_data())
+s = samples[1]
+llm = lambda ctx: ask_llm(ctx, model, temperature=0, context_length=16384, cost_callback=cost_callback)
```
JosephDavidsonKSWH, line 13: Isn't gpt-3.5-turbo only 4k context? Won't this break?

JosephDavidsonKSWH: Ah right its 16k, I see below

datasets/chapterbreak.py
```diff
@@ -38,42 +39,72 @@ def split_in_pages(text: str, max_tokens_per_split: int) -> list[str]:
...
         path = get_gdrive_file(self.name, GDRIVE_8K_ID, filename)
         with open(path) as fd:
             return json.load(fd)
 
+    def apply_sample_selection(self, samples: dict) -> dict:
+        with open(get_data_path(self.name, "chapterbreak-goodai-selection.json")) as fd:
+            selection = json.load(fd)
+        sample_selection = dict()
+        for sel in selection:
+            if not sel["solvable"]:
```
JosephDavidsonKSWH, line 84: We are not selecting unsolvable samples. So Why are they in the split file?

dev_scripts/cbtest.py
dcasbol: Yeah, at some point they unified the nomenclature, so older versions have a 4k context and the newest ones 16k.

datasets/chapterbreak.py
dcasbol: I decided to keep them to have a record of what samples have been subjected to human scrutiny, in order to save efforts in the future.

JosephDavidsonKSWH (approved):

PR #15, opened 2024-02-22
Draft: Stateful agents

JosephDavidsonKSWH:
Agents are now stateful. By default they are saved in  `data/persistence/` and their particular form depends on the agent. `save` and `load` are abstract methods.

Some agents are too out of our control to save or load. So we need to handle them gracefully.

### DIFF
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -6,6 +6,8 @@ import browser_cookie3
 from model_interfaces.interface import ChatSession
 import requests
 
+from utils.constants import ResetPolicy
+
 
 def try_extract_session_cookie(cj):
     user_name = ""
@@ -27,6 +29,7 @@ class CharlieMnemonic(ChatSession):
     token: str = ""
     user_name: str = ""
     initial_costs_usd: float = 0.0
+    reset_policy: ResetPolicy = ResetPolicy.SOFT
 
     @property
     def name(self):
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -135,3 +135,10 @@ class CharlieMnemonic(ChatSession):
         )
 
         self.context = []
+
+
+    def load(self):
+        pass
+
+    def save(self):
+        pass
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -29,7 +29,6 @@ class CharlieMnemonic(ChatSession):
     token: str = ""
     user_name: str = ""
     initial_costs_usd: float = 0.0
-    reset_policy: ResetPolicy = ResetPolicy.SOFT
 
     @property
     def name(self):
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -35,6 +35,7 @@ class CharlieMnemonic(ChatSession):
         return f"{super().name} - {self.max_prompt_size}"
 
     def __post_init__(self):
+        super().__post_init__()
         browsers = [
             browser_cookie3.chrome,
             browser_cookie3.firefox,
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -139,9 +139,10 @@ class CharlieMnemonic(ChatSession):
 
         self.context = []
 
-
     def load(self):
+        # Charlie mnemonic is web based and so doesn't need to be manually told to resume a conversation
         pass
 
     def save(self):
+        # Charlie mnemonic is web based and so doesn't need to be manually told to persist
         pass
--- a/model_interfaces/interface.py
+++ b/model_interfaces/interface.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from typing import Tuple, Callable
 from dataclasses import dataclass, field
 
-from utils.constants import PERSISTENCE_DIR
+from utils.constants import PERSISTENCE_DIR, ResetPolicy
 
 
 @dataclass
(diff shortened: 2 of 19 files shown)


model_interfaces/charlie_interface.py
```diff
@@ -135,3 +138,10 @@ def reset(self):
         )
 
         self.context = []
+
+
+    def load(self):
+        pass
+
+    def save(self):
```
dcasbol, line 146: I believe this is the right implementation, because Charlie works through an online API, but it would be good to have that explained in a comment.

model_interfaces/interface.py
```diff
@@ -27,6 +31,14 @@ def message_to_agent(self, user_message: str) -> Tuple[str, datetime, datetime]:
     def name(self):
         return self.__class__.__name__
 
+    @property
+    def save_path(self):
+        return PERSISTENCE_DIR.joinpath(f"{self.dataset_name} - {self.name}")
+
+    @property
+    def save_name(self):
+        return f"{self.dataset_name} - {self.name}"
```
dcasbol, line 40: This doesn't seem right. The unique agent name should be `<run_name> - <self.name>`. Also, `save_name` can be used in `save_path`.

model_interfaces/ltm_agent_1.py
```diff
@@ -313,6 +318,32 @@ def reset_all(self):
...
+            message_hist.append(Message(**m))
+        self.message_history = message_hist
+
+        fname = PERSISTENCE_DIR.joinpath(self.save_name + "_user_info.json")
+        with open(fname, "r") as fd:
+            self.user_info = json.load(fd)
+
+        fname = PERSISTENCE_DIR.joinpath(self.save_name + "_mem.json")
+        with open(fname, "r") as fd:
+            self.text_mem.set_state(json.load(fd))
```
dcasbol, line 346: Since the base class already provides `self.save_path` which is a directory, we could save those files in there e.g.
```python
fname = self.save_path.joinpath("mem.json")
```

runner/run_benchmark.py
```diff
@@ -138,10 +140,13 @@ def check_result_files(run_name: str, agent_name: str, force_removal: bool = Fal
                     default_yes=False,
                 ):
                     colour_print("red", "Run aborted.")
-                    return
+                    exit()
```
dcasbol, line 143: Why is this an exit now? Seems too drastic.

JosephDavidsonKSWH: I have no idea how it worked before. If you choose not to continue, the run continues anyway. The run actually aborts now.

runner/run_benchmark.py
```diff
@@ -175,10 +180,14 @@ def _main(configuration: str, agent_name: str, max_prompt_size: Optional[int], y
         logging.warning("Running without a maximum prompt size.")
     else:
         print(f"Maximum prompt size: {max_prompt_size}")
-    agent = get_chat_session(agent_name, max_prompt_size=max_prompt_size)
+
+    agent = get_chat_session(agent_name, max_prompt_size=max_prompt_size, dataset_name=config['run_name'])
```
dcasbol, line 184: Alright, I see that you're using the right value, just the variables have the wrong name. It should be `run_name` instead of `dataset_name` all the way down to the agent interface.

JosephDavidsonKSWH: I can refactor that!

runner/scheduler.py
```diff
@@ -258,23 +258,29 @@ def setup_iterator(self, test_group):
             self.travel_to_dt(self.master_log.log[-1].timestamp)
```
dcasbol, line 258: I don't think it matters from a functional standpoint, but maybe it will confuse the user? If the reset policy is hard, then this timestamp will probably correspond to a test that will be reset.

runner/scheduler.py
```diff
@@ -258,23 +258,29 @@ def setup_iterator(self, test_group):
...
 
         # Fast forward examples to the last action that was run
-        for run_id in test_actions_taken.keys():
-            example = tests[run_id]
-            actions_to_ff = test_actions_taken[run_id]
-            for _ in range(actions_to_ff):
-                action = example.step()
-                if isinstance(action, SendAndRegisterAction):
-                    self.register_callback(example)
+        # If the reset priority for the model is soft, then fast forward the tests, otherwise all tests in progress are discarded
```
dcasbol, line 261: I got a bit confused by this comment, I'd rephrase it to "all messages from partial tests are discarded".

model_interfaces/interface.py
```diff
@@ -3,13 +3,17 @@
...
+
 
 @dataclass
 class ChatSession(ABC):
     history: list[dict[str, str | datetime]] = field(default_factory=list)
     costs_usd: float = 0
     is_local: bool = False
     max_message_size: int = 1000
+    dataset_name: str = ""
+    reset_policy: ResetPolicy = ResetPolicy.HARD
```
dcasbol, line 16: What about making the reset policy soft by default, since that's the one that will be used mostly? To implement an agent one has to implement `save` and `load` anyway, so setting the reset policy could be part of it.

runner/run_benchmark.py
dcasbol: And what about raising an exception instead? I personally try not to use `exit` unless it's a quick script.

runner/scheduler.py
JosephDavidsonKSWH: I don't want to edit the log after a hard reset by removing events that did happen and contributed to tokens/money spent. So rather than a complicated time resolution process, setting it to the last event (whether kept or discarded) is simple enough.

model_interfaces/memgpt_interface.py
```diff
@@ -224,6 +227,13 @@ def get_sent_messages(self, message_list):
 
         return message_strings
 
+    def save(self):
+        colour_print("CYAN", "MemGPT will save in your home directory under '.memgpt'")
```
dcasbol, line 231: Move this to the init method.

dcasbol (approved):

PR #19, opened 2024-02-27
LTM-82 - Restaurant task

dcasbol:


datasets/restaurant.py
```diff
@@ -0,0 +1,232 @@
...
+    score: int = 0
+    max_score: int = 1
+    filler_tokens: int = 1000
+    expected_responses: list[str] = field(default_factory=list)
+    script: List[str] = field(default_factory=lambda: [""])  # Required for compatibility, but no actual script.
+
+    def __post_init__(self):
+        super().__post_init__()
+        assert self.cost_callback is not None, "Dynamic examples require a cost callback."
+        assert self.evaluation_fn is None, "Dynamic examples have their own evaluation function."
```
JosephDavidsonKSWH, line 25: These asserts seem to be generic, but the class they are in is specific. Do you want to try and make a `DyanmicTextExample` class?

datasets/restaurant.py
```diff
@@ -0,0 +1,232 @@
...
+from dataclasses import dataclass, field
+from dataset_interfaces.interface import DatasetInterface, TestExample, TestAction, SendMessageAction, WaitAction
+
+from utils.openai import ask_llm, LLMContext, make_user_message
+from goodai.helpers.json_helper import sanitize_and_parse_json
+
+
+@dataclass
+class RestaurantExample(TestExample):
+    cost_callback: Callable[[float], None] = None
```
JosephDavidsonKSWH, line 14: The cost_callback is a member of the `DatasetInterface` class so you might be able to use that?

datasets/restaurant.py
dcasbol: oh yeah, definitely. That's done but not pushed yet.

datasets/restaurant.py
dcasbol: Yes. `DynamicExample` objects have it because they might perform calls to gpt-3 to evaluate things or guide the script. The cost is tiny, but this way it's accounted for.

dataset_interfaces/interface.py
```diff
@@ -93,6 +103,10 @@ def get_path(self, run_name: str) -> Path:
         return make_testdef_path(run_name, self.dataset_name, self.example_id)
 
     def __post_init__(self):
+        assert self.dataset_generator is not None
+        if self.evaluation_fn is None:
+            self.evaluation_fn = self.dataset_generator.evaluate_correct
```
JosephDavidsonKSWH, line 108: If you are doing some `TestExample` attributes as properties, you can also add this/

JosephDavidsonKSWH (approved):

PR #27, opened 2024-03-07
upload gpt-4 usages to newest version

dcasbol:


JosephDavidsonKSWH (approved):
We can make use of `gpt-4-turbo-preview` that points at the latest, but that can be done later along with the big refactor.

PR #30, opened 2024-03-08
revamped memgpt + interface

JosephDavidsonKSWH:


runner/run_benchmark.py
```diff
@@ -49,10 +49,8 @@ def get_chat_session(name: str, max_prompt_size: Optional[int], run_name: str) -
         model = name.removesuffix("ts-") + "-preview"
         return TimestampGPTChatSession(model=model, **kwargs)
     elif name == "memgpt":
-        if max_prompt_size is not None:
-            return MemGPTChatSession(_max_prompt_size=max_prompt_size, run_name=run_name)
-        else:
-            return MemGPTChatSession(run_name=run_name)
+        return MemGPTChatSession(run_name=run_name)
```
dcasbol, line 52: so we're not messing up with the `max_prompt` anymore, right?

dcasbol (approved):

runner/run_benchmark.py
JosephDavidsonKSWH: Not in the way that we did before. That was a little janky, but this new method seems to work much better.

PR #31, opened 2024-03-11
Trigger response eval fix

dcasbol:
We should run `dev_scripts/migr-trigger_response_evals.py` on those runs and agents that did not have this evaluation available at benchmark time.


datasets/trigger_response.py
```diff
@@ -75,21 +85,40 @@ def generate_examples(self, num_examples):
...
+        try:
+            eval_json = sanitize_and_parse_json(eval_str)
+            present = eval_json["present"]
+            if present:
+                present = rouge_score(expected, eval_json["sentence"]) > 0.75
+        except (ValueError, JSONDecodeError, KeyError) as exc:
+            return 0, f"Could not evaluate due to a JSON parsing error: {repr(exc)}"
+        not_str = "" if present else "not "
+        return int(present), f"'{expected}' is {not_str}in the response."
+
```
JosephDavidsonKSWH, line 102: Given the conversation for `TriggerResponse`, can we just string match here?

dcasbol: I almost say "yes", but then I remembered about our conversation last week, about us constantly switching back and forth between string matching and GPT evaluations 😛 The thing is, agents sometimes stick the user's name in the middle of the sentence, keeping the rest verbatim. You could get around that with rouge-L scores, but sentence lengths affect it quite a bit, and agents usually say more than the sentence alone. So that's why the combination of a loose and somewhat flexible sequence extraction (GPT) + the high rouge score on that sentence.

PR #32, opened 2024-03-11
some fixes for claudes context trimming

JosephDavidsonKSWH:


model_interfaces/claude_interface.py
```diff
@@ -53,6 +55,23 @@ def reply(self, user_message: str) -> str:
...
+                to_be_added = message
+                continue
+
+            if message["role"] != to_be_added["role"]:
+                messages.append(to_be_added)
+
+            to_be_added = message
+
+        messages.append(to_be_added)
+        return messages
```
dcasbol, line 73: Does this simply discard messages if +1 messages with the same role are adjacent in the context? Any hint on why this happens?

JosephDavidsonKSWH: Yes, so I reused the chatgpt version for trimming the context, because the structure is the same. But the ChatGPT version has a system prompt, and the context trimmin assumes there is one there. It keeps the first message.

In this case, the first message is a user message about the chapterbreak task. when we remove messages, we take them from `context[1:]` and that is an alternating `user, assistant, user, assistant, etc`. And if `context[1]` after trimming is a `user` message, then the there are two in a row and that causes the error. I could alternatively try a more surgical check of just context[0] and context[1] and maybe put a system message in?.

dcasbol: That would be much better, yeah... mainly to avoid silent errors and be up to what's going on in the system. A few assertions could be a good complement to it.

PR #37, opened 2024-03-18
LTM-107 - Resuming dynamic tests

dcasbol:
Resuming process has been changed. Now the system follows the master log and goes step by step, keeping both the tests and the waiting list up to date.

### DIFF
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from typing import List, Callable, Tuple, Optional, Any, Iterator, Dict
+from runner.master_log import MasterLog
 
 import tiktoken
 from goodai.helpers.json_helper import sanitize_and_parse_json
@@ -207,9 +208,11 @@ class DynamicExample(TestExample):
     max_score: int = 0
     expected_responses: list[str] = field(default_factory=list)
     reasoning: list[str] = field(default_factory=list)
-    script: List[str] = field(default_factory=lambda: [])  # Updated dynamically by `say`
+    script: List[str] = field(default_factory=lambda: [])  # Updated dynamically by `say` method
     action: SendMessageAction = None  # Keeps the last SendMessageAction
     wait = WaitAction
+    llm_call_idx: int = -1
+    master_log: MasterLog = None  # Set by runner to cache llm calls
 
     @property
     def evaluation_fn(self) -> Callable[[List[str], list[str], List[Any]], tuple[int, int, List[str]]]:
@@ -223,7 +226,13 @@ class DynamicExample(TestExample):
         return self.score, self.max_score, self.reasoning
 
     def ask_llm(self, context: LLMContext, **kwargs) -> str:
-        return self.dataset_generator.ask_llm(context, **kwargs)
+        self.llm_call_idx += 1
+        response = self.master_log.get_cached_response(self.unique_id, self.llm_call_idx)
+        if response is not None:
+            return response
+        response = self.dataset_generator.ask_llm(context, **kwargs)
+        self.master_log.add_llm_call(self.unique_id, datetime.now(), response)
+        return response
 
     @abstractmethod
     def action_iter(self) -> Iterator[TestAction]:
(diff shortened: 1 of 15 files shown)


datasets/jokes.py
```diff
@@ -54,7 +54,10 @@ def generate_examples(self, num_examples):
                 selected_jokes.append(joke)
                 is_question.append(False)
                 time_jump = create_time_jump(self.minutes_low, self.minutes_high)
-                waits.append(WaitCreator.create_wait(tokens=randint(self.filler_tokens_low, self.filler_tokens_high), time=time_jump))
+                waits.append(WaitCreator.create_wait(
+                    tokens=randint(self.filler_tokens_low, self.filler_tokens_high),
```
JosephDavidsonKSWH, line 58: Should we remove the random filler? I don't think we have ever used it

dcasbol: Yeah, let's take the chance.

dataset_interfaces/interface.py
```diff
@@ -132,7 +140,7 @@ def action_iter(self) -> Iterator[TestAction]:
                 yield SendAndRegisterAction(msg, is_question=is_q)
             else:
                 yield SendMessageAction(msg, is_question=is_q)
-            if any(wait.values()):
+            if len(wait) > 0:
```
JosephDavidsonKSWH, line 143: Is this correct? The create_wait function will return by default this:  `{"tokens": None, "time": None, "percentage_finished": None}`

datasets/trigger_response.py
```diff
@@ -55,6 +56,7 @@ class TriggerResponseDataset(DatasetInterface):
     description: str = "Tell the agent to respond in a particular way when a trigger is given. Test the agent."
     trigger_activations: int = 3
     reset_message: str = "Cancel any instructions as to what sentence you should say whenever I do something in particular."
+    rouge_score_threshold: float = 0.75
```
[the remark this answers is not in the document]
dcasbol: I'd say it's quite solid. I tuned it more or less based on some comparisons I made in my machine, although I reckon it might be tuned further to tighten or loosen up the constraint. The ROUGE-L score is a value between 0 (very different) and 1 (equal), but it's very dependent on the length of the sequences and it can decrease very rapidly if the sentences don't have similar lengths.

dcasbol:
I have added a template for the `Benchmark 3` configuration, to accomodate the changes in the config. We can then update it on every PR until we reach a final state.

runner/scheduler.py
```diff
@@ -241,48 +245,72 @@ def pick_next_test_id(self, tests: dict[str, TestExample]) -> str:
...
+            self.wait_list.pop(evt.test_id, None)  # Since the test is performing an action, it can't be waiting.
+            match evt.type:
+                case EventType.BEGIN:
+                    result, skip = self.initialise_result(test)
+                    assert not skip
+                    self.in_progress_results[evt.test_id] = result
+                case EventType.SEND_MESSAGE:
+                    action = test.step()
+                    assert isinstance(action, SendMessageAction)
+                    assert action.message == evt.data["message"]
```
JosephDavidsonKSWH, line 276: This is exactly the problem that the Restaurant task had when resuming. This assert will likely fail on loading

dcasbol: Good catch. In this case, responses are taken from the log, so the generator function has fewer chances to drift away from what was logged, but it could happen... Funny, it did not happen once in my tests, so I even forgot about the possibility. I'll go through it again and see what I can do. I remember thinking about implementing a sort of "forced mode" in the dynamic tests.

dataset_interfaces/interface.py
JosephDavidsonKSWH: Oh, I missed that `if not None` at the end of  line 74

runner/scheduler.py
dcasbol: Ideally, GPT would be deterministic and I wouldn't have to worry about these things, because any randomness in the dataset is also seeded 😅

dcasbol: Actually, I think I answered myself. I will see if I can log the calls to `ask` from dynamic tests, so it can be loaded too.

JosephDavidsonKSWH (approved):

PR #38, opened 2024-03-20
LTM-106 - Locations Directions evaluations

dcasbol:
The task is evaluated by actually following the agent's directions. The process is as follows:
1. GPT-3 converts the agent's response into a structured JSON.
2. We follow both the exact original route and the agent's directions.
3. We compare that the final position is the same.

The dataset seeding has been also improved.


datasets/locations_directions.py
```diff
@@ -1,28 +1,117 @@
...
         # Now the answer should include all the directions and locations
-        answer = ""
-        for l in location_info[1:-1]:
-            answer += f"From {l[1]}, go {l[3]}km {l[2]} to {l[0]}.\n"
+        pos = [0, 0]
+        directions = ["(Just an example, if exact instructions are followed)"]
+        for destination, origin, direction, distance in location_info[1:-1]:
+            axis = 0 if direction.lower() in ["west", "east"] else 1
+            sign = 1 if direction.lower() in ["north", "east"] else -1
+            pos[axis] += sign * distance
```
dcasbol, line 36: I left this in by mistake. Removing it...

datasets/locations_directions.py
```diff
@@ -1,28 +1,116 @@
...
+            directions.append(f"From {origin}, go {distance}km {direction} to {destination}.")
 
-        answer += f"From {location_info[-1][1]}, go {last_move[1]}km {last_move[0]} to {location_info[-1][0]}."
+        directions.append(f"From {location_info[-1][1]}, go {last_move[1]}km {last_move[0]} to {location_info[-1][0]}.")
+        answer = "\n".join(directions)
 
         return last_move, [answer]
+
+    def parse_directions(self, expected_answer: str) -> list[DirectionDict]:
+        pattern = r"^From (?P<origin>\w+(?: \w+){0,2}), go (?P<distance>\d)km (?P<direction>\w+) to (?P<destination>\w+(?: \w+){0,2}).$"
```
JosephDavidsonKSWH, line 41: AAAAAAAAAAAAAAAAAAAAAAAAAAAAA

dcasbol: :joy: Some hints:
- If it's between brackets, it will extract it into a group.
- `(?P<name>...)` Means that the group will have the key `name` (I looked it up in stackoverflow).
- `(?:...)` Means that it shouldn't extract this i.e. the brackets are just for grouping it in the expression.

PR #39, opened 2024-03-20
Spy fix

JosephDavidsonKSWH:
Small temp fix to give credit to the agents that actually do recall the correct messages


dcasbol:
It looks very much ad-hoc, but I guess it works for a temp fix. However, these changes must have altered the overall scores, so the comparison reports should be re-generated too.

PR #40, opened 2024-03-20
LTM-105 - Irregular max scores

dcasbol:
The only task with irregular max scores was _Shopping List_. The task now scores a max of `item_changes`. The punctuation can be broken down in thirds:
1. One third depends on the number of expected items present.
2. Another third depends on those items' quantities matching the expected values.
3. The last third is given if there are no hallucinated items.

### DIFF
--- a/datasets/shopping.py
+++ b/datasets/shopping.py
@@ -153,8 +153,8 @@ class ShoppingDataset(DatasetInterface):
             for item in answer_items:
                 assert isinstance(item["item"], str)
                 assert isinstance(item["quantity"], int)
-        except (JSONDecodeError, ValueError, KeyError, AssertionError):
-            msg = f"Response not in correct format:\n{responses[0]}"
+        except (JSONDecodeError, ValueError, KeyError, AssertionError) as exc:
+            msg = f"Response not in correct format ({repr(exc)}):\n{responses[0]}"
             logging.exception(msg)
             return score, max_score, [msg]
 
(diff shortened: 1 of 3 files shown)


datasets/shopping.py
```diff
@@ -125,41 +125,68 @@ def evaluate_correct(
...
-        num_correct = max(num_correct - penalties, 0)
-        if num_correct == max_score:
-            reasoning = "All items and quantities match."
-        else:
-            reasoning = "\n".join(errors)
-        return num_correct, max_score, [reasoning]
+                assert isinstance(item["item"], str)
+                assert isinstance(item["quantity"], int)
+        except (JSONDecodeError, ValueError, KeyError, AssertionError):
+            msg = f"Response not in correct format:\n{responses[0]}"
```
dcasbol, line 157: I'll add the exception to this message to add information.

datasets/shopping.py
```diff
@@ -125,41 +125,68 @@ def evaluate_correct(
...
+            if not matched:
+                hallucinated.append(name)
+                continue
+            present.append(name)
+            if item["quantity"] == expected_items[key]:
+                num_correct += 1
+            else:
+                reasoning.append(f"Wrong quantity for {name}: {item['quantity']} vs {expected_items[key]}.")
+
+        score += len(present) / len(expected_answers)
```
JosephDavidsonKSWH, line 174: This could theoretically lead to an unbounded score if I hallucinate a lot of items. Maybe constrain it to 1?

JosephDavidsonKSWH: Actually not since the addition to present is after the hallucination check. Maybe rename the variable to `present_and_correct` or something?

JosephDavidsonKSWH (approved):

PR #42, opened 2024-03-25
LTM-104 - Repeated samples

dcasbol:
All datasets use now `self.random` as the random number generator, which is seeded automatically when `generate_examples` is called.

### DIFF
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -265,6 +265,12 @@ class DatasetInterface(ABC):
     def __post_init__(self):
         self.random = Random(self.seed)
 
+    def __getattribute__(self, item):
+        # Force setting the seed before generating samples
+        if item == "generate_examples":
+            self.random = Random(self.seed)
+        return super().__getattribute__(item)
+
     @property
     def data_path(self) -> Path:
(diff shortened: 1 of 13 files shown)


dataset_interfaces/interface.py
```diff
@@ -265,6 +265,12 @@ class DatasetInterface(ABC):
     def __post_init__(self):
         self.random = Random(self.seed)
 
+    def __getattribute__(self, item):
```
JosephDavidsonKSWH, line 268: In what situation is this needed? `__post_init__` should take care of the setting of the seed, no?

dcasbol: I guess it depends on the behaviour we want. I'd expect `generate_examples` to be called only once per dataset, or at least be expected to generate the same samples if it's called twice.

JosephDavidsonKSWH: Hmmm, I see your point. I guess we should expect it to generate the same tests each call to `generate_examples`. In this case I would remove the mention of the random from `__post_init__` if that won't cause any problems.

JosephDavidsonKSWH (approved):

PR #43, opened 2024-03-25
Ltm 118: Spy meeting fixes

JosephDavidsonKSWH:


dcasbol:
Some points:
1. What's the main motivation for these fixes?
2. There's some code commented out, which seems for testing.

JosephDavidsonKSWH:
1.) Initially in the tests, there was `"bring a fast land escape vehicle."` Which the agents never interpreted as something that needed "decoded". So I rephrased that and tested them all again, doing that I found some other cases where the LLM phrases something that I haven't string matched. So I did those fixes too.    

2.) Yeah I ran every singe combination as a test. I can remove it.

dcasbol (approved):

PR #45, opened 2024-03-26
LTM-95 - Simplify OpenAI models

dcasbol:
- We now have a very small but exhaustive list of supported models.
- All context and cost information is in the same place.
- There are a couple of aliases and constants, explicitly defined.
- Construction of agent in `run_benchmark` has been simplified.
- `utils.openai` has been renamed to `utils.llm`.

### DIFF
--- a/runner/run_benchmark.py
+++ b/runner/run_benchmark.py
@@ -39,11 +39,11 @@ def get_chat_session(name: str, max_prompt_size: Optional[int], run_name: str) -
         return MemGPTChatSession(run_name=run_name)
     if name.startswith("langchain_"):
         suffix = "_".join(name.split("_")[1:])
-        mem_type = dict(
-            sb_a=LangchainMemType.SUMMARY_BUFFER,
-            kg_a=LangchainMemType.KG,
-            ce_a=LangchainMemType.CONVERSATION_ENTITY,
-        ).get(suffix, None)
+        mem_type = {
+            "sb_a": LangchainMemType.SUMMARY_BUFFER,
+            "kg_a": LangchainMemType.KG,
+            "ce_a": LangchainMemType.CONVERSATION_ENTITY,
+        }.get(suffix, None)
         if mem_type is None:
             raise ValueError(f"Unrecognized LangChain memory type {repr(suffix)}.")
         return LangchainAgent(model_name="gpt-3.5-turbo-instruct", mem_type=mem_type, **kwargs)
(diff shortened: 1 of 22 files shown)


runner/run_benchmark.py
```diff
@@ -25,61 +25,60 @@
...
+        if mem_type is None:
+            raise ValueError(f"Unrecognized LangChain memory type {repr(suffix)}.")
+        return LangchainAgent(model_name="gpt-3.5-turbo-instruct", mem_type=mem_type, **kwargs)
+    if name.startswith("ltm_agent_"):
+        agent_nr = "_".join(name.split("_")[2:])
+        variant = {
+            "1": LTMAgentVariant.QG_JSON_USER_INFO,
+            "2": LTMAgentVariant.SEMANTIC_ONLY,
+            "3": LTMAgentVariant.TEXT_SCRATCHPAD,
+        }.get(agent_nr, None)
```
JosephDavidsonKSWH, line 56: This form is inconsistent with the one above. Pick a style of creating a dict and use it universally.

dcasbol: I do prefer using `dict` because it looks more pythonic, like when you call a function with keyword arguments, but it's not always possible. If the elements contain special characters or start with a digit, then I have no other option than using the string-based approach. I can stick to the dict literal in this case though, if you think it's for the best.

runner/run_benchmark.py
```diff
@@ -25,61 +25,60 @@
...
+    if name == "human":
         return HumanChatSession(**kwargs)
-    else:
-        raise ValueError(f"Unrecognized agent: {name}")
+
+    try:
+        model = get_model(name.removesuffix("ts-"))
+        if model.startswith("claude-"):
+            assert not name.startswith("ts-")
+            cls = ClaudeChatSession
```
JosephDavidsonKSWH, line 72: Why is this check for Claude in the `ts` section?. And also, why not just check for `startswith("ts")` and branch from there?

dcasbol: The timestamped interface currently supports only OpenAI models.

JosephDavidsonKSWH: So this is a defence against someone doing `ts-claude`? I think that is unessecary since we know that it will fail immediately afterward. Would it not be better to just have a match for the name on `claude`, then do the whole `ts` parsing for the timestamped bit. It feels a little convoluted that  in order to select a claude model, you have to make sure that its not some other model first. 

Also just noticed, the function `removesuffix` should be `removeprefix`

runner/run_benchmark.py
JosephDavidsonKSWH: Ah I see, okay then I think its okay. it was just confusing to see both styles so close to each other without knowing why.

dcasbol: I agree in that it's confusing, being so close to each other.

JosephDavidsonKSWH (approved):

PR #46, opened 2024-04-02
LTM-131 - Benchmark intro

dcasbol:


runner/scheduler.py
```diff
@@ -355,6 +355,15 @@ def run_tests(self):
...
         percentage_per_test = 100 / len(self.tests)
+
+        # Introduce the benchmark, if running from the start.
+        if len(self.master_log.log) == 0:
+            self.send_message("", SendMessageAction(message=(
+                "I am going to subject you to a Long-Term Memory benchmark. In the following, I will be giving you "
+                "different kinds of information and I expect you to answer extremely briefly, only providing the "
+                "responses that you are required to provide. Otherwise, provide just short confirmations. Understood?"
+            )))
+
```
JosephDavidsonKSWH, line 366: Do you have an idea of how this grounding might change the overall quality of the responses?

JosephDavidsonKSWH (approved):

runner/scheduler.py
dcasbol: I believe it will improve the responses in general. It did improve gpt-3.5's, and it was similar to what I had to do when I asked my brother to take the test 😅

PR #47, opened 2024-04-02
LTM-116 - Keep count of time and tokens wrt the agent

dcasbol:
It doesn't make much sense to count the overall benchmark duration and put it on the final agent's records. In doing it that way, we are including our system's inefficiencies in the agent's statistics.

Tokens are also counted separatedly now. Agent's tokens are used for computing the `avg_tokens_per_second`, instead of having a moving average (which is used nowhere anyway), and also to compute the benchmark speed (tests per hour).

### DIFF
--- a/runner/scheduler.py
+++ b/runner/scheduler.py
@@ -54,11 +54,10 @@ class TestRunner:
     in_progress_results: dict[str, TestResult] = field(default_factory=dict)
     traveller: Optional[time_machine.travel] = None
     wait_list: dict[str, dict[str, int | datetime]] = field(default_factory=dict)
-    current_token_count: int = 0
-    avg_tokens_per_second: Optional[float] = None
+    agent_token_count: int = 0
+    total_token_count: int = 0
     test_managing_costs_usd: float = 0
-    benchmark_duration_seconds: float = 0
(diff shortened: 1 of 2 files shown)


runner/scheduler.py
```diff
@@ -54,11 +54,10 @@ class TestRunner:
     in_progress_results: dict[str, TestResult] = field(default_factory=dict)
     traveller: Optional[time_machine.travel] = None
     wait_list: dict[str, dict[str, int | datetime]] = field(default_factory=dict)
-    current_token_count: int = 0
-    avg_tokens_per_second: Optional[float] = None
+    agent_token_count: int = 0
```
JosephDavidsonKSWH, line 57: So `agent_token_count` is only for tokens generated by the agent and is only used to calculate the tokens per second and speed of the agent?

JosephDavidsonKSWH (approved):

runner/scheduler.py
dcasbol: That's right. And we haven't used that metric so far, although it seems useful. We do use the _tests per hour_ metric, which I guess it's an equivalent. We could also maybe use `avg_tokens_per_second` for estimations, but I'm not sure which.

PR #49, opened 2024-04-05
Ltm 124 big and small

JosephDavidsonKSWH:
Big and small numbers:

- Big numbers now guarantees that the memory span of a test will be less than the big number.
- Filler token values are removed, waits are calculated on either a 75/25 context split, or a 50/50 split if `number_of_questions > 1`
- Setup statements are uniformly distributed across the first number in the split, questions are uniformly distributed across 90% of the second number
- When calculating filler, we now also take into account the length of the probable response (which is a list of the answers). This means that the filler consists of the trivia + agent response and that length is more true to the number of filler tokens we desire.
- Percentage waiting has been removed - it doesn't make sense in the context of the big number.
- Nomenclature in reports has been changed and terms related to memory span have been cleared up a little.
- On creating waits, we have assertions to filter out datasets that have more data than the memory span.
- Waits are now the responsibility of the `TestExample`, not of the `DatasetGenerator` This was done to try and reduce the dependency of `TestExample` initialisation on the `DatasetGenerator`. What was happening is that The DatasetGenerator was creating the TestExample, which then called the DatasetGenerator in `__post_init__`. Which was strange and added odd dependencies on information in the `DatasetGenerator` that would also be needed at load time, but we don't save the generators, so it was all very complicated.  
- `answer_statement_idx` is gone, the memory span is calculated from the first statement to the last.
- Detailed reports now show how many needles are in a test.

### DIFF
--- a/dataset_interfaces/factory.py
+++ b/dataset_interfaces/factory.py
@@ -21,6 +21,8 @@ from datasets.chapterbreak import ChapterBreakDataset
 from datasets.restaurant import RestaurantDataset
 from copy import deepcopy
 
+from utils.files import parse_definition_path
+
 DATASETS = {
     "names": NamesDataset,
     "colours": ColourDataset,
@@ -68,8 +70,7 @@ class DatasetFactory:
         args = deepcopy(run_configuration["datasets"]["args"])
 
         # Get the name of the dataset
-
-        path_dataset_name = test_example_path.split(os.sep)[-2]
+        path_dataset_name = parse_definition_path(test_example_path)["dataset_name"]
         dataset = DATASETS_BY_NAME.get(path_dataset_name, None)
 
         if dataset is None:
--- a/dataset_interfaces/factory.py
+++ b/dataset_interfaces/factory.py
@@ -1,4 +1,6 @@
-from dataset_interfaces.interface import TestExample
+import os
+
+from dataset_interfaces.interface import TestExample, DatasetInterface
 from datasets.conficting_personal_information import ConflictingPersonalInformationDataset
 from datasets.delayed_recall import DelayedRecallDataset
 from datasets.how_to_think import HowToThinkDataset
@@ -60,3 +62,33 @@ class DatasetFactory:
             if example.example_id == "":
                 example.example_id = str(i)
         return examples
+
+    @staticmethod
+    def create_dataset_for_example(run_configuration: dict,  test_example_path: str) -> DatasetInterface:
+        args = deepcopy(run_configuration["datasets"]["args"])
+
+        # Get the name of the dataset
+
+        path_dataset_name = test_example_path.split(os.sep)[-2]
+        dataset = DATASETS_BY_NAME.get(path_dataset_name, None)
+
+        if dataset is None:
+            raise ValueError(f"No dataset could be resolved from TestExample path: {test_example_path}. Tried {path_dataset_name}")
+
+        # Use the class to get the 'config name'
+        config_name = ""
+        for name, ds in DATASETS.items():
+            if ds == dataset:
+                config_name = name
+                break
+
+        # Use config name to get any extra config from the run configuration
+        extra_args = {}
+        for dataset_config in run_configuration["datasets"]["datasets"]:
+            if dataset_config["name"] == config_name:
+                extra_args = dataset_config.get("args", {})
+                break
+
+        args.update(extra_args)
+        del args["dataset_examples"]
+        return dataset(**args)
--- a/dataset_interfaces/gpt_generated.py
+++ b/dataset_interfaces/gpt_generated.py
@@ -88,7 +88,3 @@ class GPTGenerated(DatasetInterface, ABC):
         self, questions: List[str], responses: List[str], expected_answers: List[Any]
     ) -> Tuple[int, int, List[str]]:
         return self.evaluate_correct_gpt(questions, responses, expected_answers)
-
-    def answer_statement_idx(self, example: TestExample) -> Tuple[int, int]:
-        # In GPT-generated tasks, the relevant info is given in the first script line.
-        return 0, 0
--- a/dataset_interfaces/gpt_generated.py
+++ b/dataset_interfaces/gpt_generated.py
@@ -77,7 +77,6 @@ class GPTGenerated(DatasetInterface, ABC):
                 expected_responses=expected_responses,
                 is_question=is_question,
                 uses_callback=self.uses_callback,
-                memory_span=self.memory_span,
             )
 
             examples.append(example)
--- a/dataset_interfaces/gpt_generated.py
+++ b/dataset_interfaces/gpt_generated.py
@@ -77,6 +77,7 @@ class GPTGenerated(DatasetInterface, ABC):
                 expected_responses=expected_responses,
                 is_question=is_question,
                 uses_callback=self.uses_callback,
+                memory_span=self.memory_span,
             )
 
             examples.append(example)
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -207,13 +207,9 @@ class TestExample:
         assert script_tokens < memory_span, f"The script for test {self.dataset_name} is too long (estimated {script_tokens} tokens) for the specified memory span of {memory_span} tokens."
 
         span_minus_script = memory_span - script_tokens
-        needle_question_proportions = (.75, .25) if self.number_of_questions == 1 else (.5, .5)
 
-        # Figure out what our spacing is going to be. We will do the 75%/25% needle distribution.
-        tokens_for_needles = math.floor(span_minus_script * needle_question_proportions[0])
-        needles = len(self.is_question) - sum(self.is_question)
-        token_wait_per_needle = tokens_for_needles // needles
-        token_wait_between_questions = math.floor(span_minus_script * .9 * needle_question_proportions[1]) // sum(self.is_question)
+        # Figure out what our spacing is going to be. Just distribute all the needles and questions across 90% of the memory span
+        token_wait = math.floor(span_minus_script * 0.9) // len(self.is_question)
 
         # Create the empty waits if there are any.
         if len(self.waits) == 0:
@@ -226,9 +222,8 @@ class TestExample:
             if f != {}:
                 continue
 
-            wait = token_wait_per_needle if idx < needles else token_wait_between_questions
-            self.waits[idx] = WaitCreator.create_wait(tokens=int(wait))
-            total_wait_tokens += wait
+            self.waits[idx] = WaitCreator.create_wait(tokens=int(token_wait))
+            total_wait_tokens += token_wait
 
         assert total_wait_tokens < span_minus_script, "Sum of task waits is higher than the determined memory span."
         self.waits[-1] = WaitCreator.create_wait()
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -67,12 +67,13 @@ class SendAndRegisterAction(SendMessageAction):
 class WaitAction(TestAction):
     tokens: int = 0
     time: timedelta = field(default_factory=timedelta)
+    percentage_finished: float = 0.0
 
 
 class WaitCreator:
     @classmethod
-    def create_wait(cls, tokens: int = None, time: timedelta = None):
-        w_dict = {"tokens": tokens, "time": time}
+    def create_wait(cls, tokens: int = None, time: timedelta = None, percentage_finished: float = None):
+        w_dict = {"tokens": tokens, "time": time, "percentage_finished": percentage_finished}
         return {k: v for k, v in w_dict.items() if v is not None}
 
     @classmethod
@@ -105,6 +106,7 @@ class TestExample:
     _iter: Iterator[TestAction] = None
     waits: List[dict] = field(default_factory=list)
     random: Random = None  # Seeded random generator
+    start_token: int = 0
 
     @property
     def dataset_name(self) -> str:
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -135,8 +135,8 @@ class TestExample:
         assert self.memory_span is not None
         self.number_of_questions = len([q for q in self.is_question if q])
         self._iter = self.action_iter()
-        self.default_waits()
         self.random = Random(self.unique_id)
+        self.default_waits()
 
     def action_iter(self) -> Iterator[TestAction]:
         scripts = [self.script, self.waits, self.is_question]
@@ -209,12 +209,13 @@ class TestExample:
         assert script_tokens < self.memory_span, f"The script for test {self.dataset_name} is too long (estimated {script_tokens} tokens) for the specified memory span of {self.memory_span} tokens."
 
         span_minus_script = self.memory_span - script_tokens
+        needle_question_proportions = (.75, .25) if self.number_of_questions == 1 else (.5, .5)
 
         # Figure out what our spacing is going to be. We will do the 75%/25% needle distribution.
-        tokens_for_needles = math.floor(span_minus_script * .75)
+        tokens_for_needles = math.floor(span_minus_script * needle_question_proportions[0])
         needles = len(self.is_question) - sum(self.is_question)
         token_wait_per_needle = tokens_for_needles // needles
-        token_wait_between_questions = math.floor(span_minus_script * .9 * .25) // sum(self.is_question)
+        token_wait_between_questions = math.floor(span_minus_script * .9 * needle_question_proportions[1]) // sum(self.is_question)
 
         # Create the empty waits if there are any.
         if len(self.waits) == 0:
@@ -233,6 +234,13 @@ class TestExample:
 
         assert total_wait_tokens < span_minus_script, "Sum of task waits is higher than the determined memory span."
         self.waits[-1] = WaitCreator.create_wait()
+
+    def setup_question_ratio(self):
+        # Return normalised values
+        non_questions = len(self.script) - self.number_of_questions
+
+        return non_questions/len(self.script), self.number_of_questions/len(self.script)
+
 
 
 @dataclass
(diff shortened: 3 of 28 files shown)


dataset_interfaces/interface.py
```diff
@@ -193,6 +196,52 @@ def load(cls, file_path: Path | str) -> "TestExample":
...
+        # TODO: All this token counting is based on OpenAIs methods, which is not Anthropics - but Aanthopic has no reliable tokeniser released so....
+        script_tokens = get_tokens_for_script(self.script)
+        assert script_tokens < self.memory_span, f"The script for test {self.dataset_name} is too long (estimated {script_tokens} tokens) for the specified memory span of {self.memory_span} tokens."
+
+        span_minus_script = self.memory_span - script_tokens
+        needle_question_proportions = (.75, .25) if self.number_of_questions == 1 else (.5, .5)
+
+        # Figure out what our spacing is going to be. We will do the 75%/25% needle distribution.
+        tokens_for_needles = math.floor(span_minus_script * needle_question_proportions[0])
+        needles = len(self.is_question) - sum(self.is_question)
```
dcasbol, line 216: This would need setting some constraints in relation to needle placement e.g. needles go first and after that there can only be questions. This is untrue for the restaurant task, which constantly interleaves needles and questions.

JosephDavidsonKSWH: Yeah, this function is pretty much incompatible with the restaurant task, and probably dynamic tasks in general

datasets/instruction_recall.py
```diff
@@ -1,6 +1,9 @@
 from pathlib import Path
 from dataclasses import dataclass
+from typing import List
+
 from dataset_interfaces.gpt_generated import GPTGenerated
+from dataset_interfaces.interface import TestExample
```
dcasbol, line 6: New imports not needed.

JosephDavidsonKSWH:
So for percentage waits referring to the `memory_span`, there are a couple of considerations:

1.) The example needs to know how many tokens have elapsed since the start of the test.
2.) On loading, the example should be able to read the master log to set the elapsed tokens. After each action, the tokens for that example should be updated.
3.) When the wait is executed, the percentage is converted to number of tokens of the memory_span,  then that value should be compared to the current elapsed tokens and the difference waited for.

That feels pretty reasonable

datasets/jokes.py
```diff
@@ -53,7 +56,7 @@ def generate_examples(self, num_examples):
                 selected_jokes.append(joke)
                 is_question.append(False)
                 time_jump = create_time_jump(self.minutes_low, self.minutes_high)
-                waits.append(WaitCreator.create_wait(tokens=self.filler_tokens, time=time_jump))
+                waits.append(WaitCreator.create_wait(tokens=filler_tokens, time=time_jump))
```
dcasbol, line 59: The semantics of waits require all conditions to be met, so here this would mean that the test should wait until `filler_tokens` have benn exchanged AND `time_jump` has passed. Time jumps are now trickier because they could violate the limits of the memory span, but maybe here we can do with just token waits.

datasets/restaurant.py
```diff
@@ -36,7 +36,7 @@ def restaurant_script_iter(self) -> Iterator[TestAction]:
             "When I talk to you as the waiter ('Waiter: what will it be sir?'), then you will reply as if you were the "
             "customer at a restaurant. Give straight answers to the questions and avoid going off script. Understood?"
         )
-        yield self.wait(percentage_finished=20)
+        yield self.wait(tokens=self.memory_span * .2)
```
dcasbol, line 39: I like this alternative. It could be extracted as `self.wait_percentage(20)` for improved ease of use.

datasets/spy_meeting.py
```diff
@@ -1,3 +1,4 @@
+import math
```
dcasbol, line 1: not used

dataset_interfaces/gpt_generated.py
```diff
@@ -77,6 +77,7 @@ def generate_examples(self, num_examples) -> List[TestExample]:
                 expected_responses=expected_responses,
                 is_question=is_question,
                 uses_callback=self.uses_callback,
+                memory_span=self.memory_span,
```
dcasbol, line 80: this could be set automatically in `TestExample.__post_init__` from the reference to the example generator, to minimize the probability of getting it wrong.

dataset_interfaces/interface.py
```diff
@@ -165,6 +167,7 @@ def to_dict(self) -> dict:
             can_be_interleaved=self.can_be_interleaved,
             is_temporal=self.is_temporal,
             uses_callback=self.uses_callback,
+            memory_span=self.memory_span,
```
dcasbol, line 170: Shouldn't this be just part of the run configuration and be set from there?

runner/scheduler.py
```diff
@@ -475,6 +459,7 @@ def update_result(
                 result.max_score = max_score
                 result.reasoning = reason
 
+        result.needles = len(example.script) - example.number_of_questions
```
dcasbol, line 462: repeated from line 318

dcasbol:
Some notes that we might want to think over and discuss:
- We cannot guarantee now that the memory span will be respected in all cases, so we should probably show a warning message when that happens. Some examples:
  - Token wait goes past the memory span.
  - Time wait ends up exceeding the max. number of tokens.
- Proportions and assumptions on waits definitely need extra thought. Specially when needles and questions are intertwined, when there are 2 questions or more, etc. Some suggestions:
  - Just spread out script lines as evenly as possible within the memory span.
  - Assume that dynamic tests will follow this rule, but raise a warning when they do not.

dataset_interfaces/interface.py
```diff
@@ -445,15 +462,13 @@ def _proxy_cost_callback(self, cost_usd: float) -> None:
     def evaluate_correct(self, *args):
         raise NotImplementedError("This method should not be called. Each test example has its own evaluation function.")
 
-    @abstractmethod
-    def answer_statement_idx(self, example: TestExample) -> Tuple[int, int]:
-        pass
-
     def create_example(self, **kwargs) -> DynamicExample:
+        if self.memory_span > 0:
+            kwargs["memory_span"] = self.memory_span
```
dcasbol, line 467: Shouldn't this be always passed on to the test example? (Or actually not needed, if the example takes it automatically from the dataset generator).

runner/scheduler.py
JosephDavidsonKSWH: I will keep this one. The update here is needed for more dynamic tests like restaurant and jokes.

datasets/restaurant.py
JosephDavidsonKSWH: This is a little dangerous, as its a wait for 20% of the memory span, not until 20% of the span has elapsed.

dcasbol: Ah, you're right. Yeah, in that case it would be nice to have a shortcut for waiting until a percentage of the memory span has elapsed 👍🏾

dataset_interfaces/interface.py
JosephDavidsonKSWH: This is one of those really messy situation arising from too much coupling between `TestExample` and `DatasetInterface`. They both depend on each other in different places. `DatasetInterface` creates `TestExample` in `create_examples`, and `TestExample` creates `DatasetInterface` in `TestExample.load()`.

This little bit here was because the `load()` path creates the `DatasetInterface` instance without a configuration, then `instance.memory_span` is 0. So I had to save that value as part of the `TestExample`.

A nicer way to do this could be in injecting the `DatasetInterface` instance to the `load()` method, so we can instantiate it correctly outside the `load()` and decouple it a bit from `TestExample`.

datasets/jokes.py
JosephDavidsonKSWH: I think its fine... can you construct an instance where its not fine? My thinking is:
- For the token component, that competes with all the other tests as normal, as other test run, this token wait decreases relative to the total tokens elapsed.
- After the token wait is done, then next search for an id will check the time blocked tests first and return them.

JosephDavidsonKSWH: Running it in the debugger, yeah the time is fast forwarded first. because it resolves the time waits independent of the tokens waits.

dcasbol: I might be overzealous about this, but if one test has a wait on time, and there are sufficient tests to run until that time mark comes, then the scheduler will run those first. In that case, the maximum memory span might be exceeded.

JosephDavidsonKSWH:
For respecting the memory span... We get an accurate measure of the span after the test has concluded, so we could report to the user then (and add in the report) when the memory_span constraint has been violated.
- Time I think is fine, see the comments I have made.

Wait proportions:
- We could do a better proportion thing, where the wait is based on the script lines. I have something semi dynamic where the split is 50/50 if there is more than 1 question. Something more advanced could assess all the lines in total and do some dynamic proportion, but I wanted it to be simple to explain.
- Dynamic tests will have to do their own waits - we don't currently have a catch all for them

dataset_interfaces/factory.py
```diff
@@ -60,3 +62,33 @@ def create_examples(dataset_config: dict, universal_args: dict, max_message_size
...
                 example.example_id = str(i)
         return examples
+
+    @staticmethod
+    def create_dataset_for_example(run_configuration: dict,  test_example_path: str) -> DatasetInterface:
+        args = deepcopy(run_configuration["datasets"]["args"])
+
+        # Get the name of the dataset
+
+        path_dataset_name = test_example_path.split(os.sep)[-2]
```
dcasbol, line 72: It would be better to use `utils.files.parse_result_path`

dcasbol: I've just noticed it's only for result files, but we could have one for example files too.

dataset_interfaces/factory.py
```diff
@@ -60,3 +62,33 @@ def create_examples(dataset_config: dict, universal_args: dict, max_message_size
...
+
+        if dataset is None:
+            raise ValueError(f"No dataset could be resolved from TestExample path: {test_example_path}. Tried {path_dataset_name}")
+
+        # Use the class to get the 'config name'
+        config_name = ""
+        for name, ds in DATASETS.items():
+            if ds == dataset:
+                config_name = name
+                break
```
dcasbol, line 83: You can also use `DATASETS_BY_NAME`

dataset_interfaces/interface.py
```diff
@@ -187,12 +189,56 @@ def load(cls, file_path: Path | str) -> "TestExample":
...
+        assert script_tokens < memory_span, f"The script for test {self.dataset_name} is too long (estimated {script_tokens} tokens) for the specified memory span of {memory_span} tokens."
+
+        span_minus_script = memory_span - script_tokens
+        needle_question_proportions = (.75, .25) if self.number_of_questions == 1 else (.5, .5)
+
+        # Figure out what our spacing is going to be. We will do the 75%/25% needle distribution.
+        tokens_for_needles = math.floor(span_minus_script * needle_question_proportions[0])
+        needles = len(self.is_question) - sum(self.is_question)
+        token_wait_per_needle = tokens_for_needles // needles
+        token_wait_between_questions = math.floor(span_minus_script * .9 * needle_question_proportions[1]) // sum(self.is_question)
```
dcasbol, line 216: I thought you were going to spread the script lines across the memory span (or 90% of it).

runner/master_log.py
```diff
@@ -193,6 +191,23 @@ def as_context(self, test_id: str = ""):
...
 
+    def tokens_before_event(self, event: LogEvent) -> int:
+        event_idx = self.event_idx(event)
+
+        tokens = 0
+        for evt in self.log[:event_idx]:
+            if evt.type in [EventType.SEND_MESSAGE, EventType.SEND_FILL, EventType.RESPONSE_MESSAGE, EventType.RESPONSE_FILL]:
+                tokens += token_len(evt.data["message"])
+
+        return tokens
```
dcasbol, line 202: I admit that this works, but maybe a bit overkill? What about storing the token count in the test start event?

runner/scheduler.py
```diff
@@ -120,19 +119,31 @@ def reset_time(self):
...
+        if action.percentage_finished > 0.0:
+            span_percent = example.dataset_generator.memory_span / 100
+            percent_as_tokens = span_percent * action.percentage_finished
+
+            # Compute at what token we are at relative to this test's memory span, then calculate the difference.
+            tokens_elapsed = self.total_token_count - example.start_token
+            token_wait = percent_as_tokens - tokens_elapsed
+
+        else:
+            token_wait = action.tokens
```
dcasbol, line 135: This might override `action.tokens` and could also result in a negative token count. It would be safer to take the `max(0, percent_as_tokens - tokens_elapsed, action.tokens)`.

runner/scheduler.py
```diff
@@ -274,11 +275,10 @@ def fast_forward_tests(self, tests: dict[str, TestExample]):
                 case EventType.WAIT:
                     action = test.step()
                     assert isinstance(action, WaitAction)
-                    assert action.tokens == evt.data["tokens"]
+                    assert action.tokens == evt.data["tokens"] or action.percentage_finished > 0.0
                     assert action.time == evt.data["time"]
-                    assert action.percentage_finished == evt.data["percentage_finished"]
                     self.travel_to_dt(evt.timestamp)
-                    self.set_to_wait(evt.test_id, action, log_this=False)
+                    self.set_to_wait(test, action, log_this=False)
```
dcasbol, line 281: Another assertion could be placed after `set_to_wait` to make sure that the tokens match (the percentage is not logged, but the number of tokens to wait).

dcasbol: I was thinking about `set_to_wait` returning a copy of the wait dict and then comparing that to `evt.data["tokens"]`. It's just a suggestion though, to make sure that everything matches during the fast-forward.

dataset_interfaces/factory.py
JosephDavidsonKSWH: No I can't. `DATASETS_BY_NAME` uses the "nice" name that is contained in `ds.name`. I need to get the name we use in the configuration files, which is only contained in `DATASETS`

dataset_interfaces/interface.py
JosephDavidsonKSWH: I thought I wasn't? I said that it worked fine as-is for 8k tokens, and with the warnings and guards we could just leave it for now and see if there is an issue.

dataset_interfaces/factory.py
dcasbol: Ah you're right. I didn't read it thoroughly and thought it was a reverse dict.

dataset_interfaces/interface.py
dcasbol: Alright then. It just seems odd to me that those proportions are so arbitrary. I think I'd like it more if `span_minus_script * 0.9` would be split in equal pieces and placed between every two consecutive script lines.

dcasbol (approved):
I reckon it's good to merge after the comments are addressed.

runner/scheduler.py
JosephDavidsonKSWH: Its a little messy, since the events in the log and the actual wait events don't line up completely. A logged wait event is an amount of time or tokens. An actual waiting event is a relative "until" time or tokens value.

```python
 def set_to_wait(self, example: TestExample, action: WaitAction, log_this: bool = True):
        unique_id = example.unique_id

        # We convert the percentage to tokens
        if action.percentage_finished > 0.0:
            span_percent = example.dataset_generator.memory_span / 100
            percent_as_tokens = math.floor(span_percent * action.percentage_finished)

            # Compute at what token we are at relative to this test's memory span, then calculate the difference.
            tokens_elapsed = self.total_token_count - example.start_token
            token_wait = max(0, percent_as_tokens - tokens_elapsed, action.tokens)

        else:
            token_wait = action.tokens

        if log_this:
            self.master_log.add_wait_event(
                unique_id,
                datetime.now(),
                tokens=token_wait,
                time=action.time,
            )
        self.wait_list[unique_id] = dict(
            tokens=self.total_token_count + token_wait,
            time=datetime.now() + action.time,
        )
```
The `wait_list` dict uses relative values, so its not comparable to the `WaitAction`, so I would have to return another new dict using the same values as those used in `add_wait_event`, which I could do, but it feels like adding clutter for the sake of the new assert.

dataset_interfaces/interface.py
```diff
@@ -187,12 +189,56 @@ def load(cls, file_path: Path | str) -> "TestExample":
...
+    def default_waits(self):
+
+        if isinstance(self, DynamicExample):
+            return
+
+        assert len(self.is_question) >= 1, "There are no questions for this test"
+        assert len(self.waits) == 0 or len(self.waits) == len(self.is_question), "Current waits should be empty or the same length as the script"
+        memory_span = self.dataset_generator.memory_span
+
+        # TODO: All this token counting is based on OpenAIs methods, which is not Anthropics - but Aanthopic has no reliable tokeniser released so....
```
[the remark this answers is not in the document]
JosephDavidsonKSWH: Not sure. It would be great to have the proper tokeniser being used everywhere. So I will create another ticket for it to investigate more thoroughly.

dataset_interfaces/interface.py
JosephDavidsonKSWH: Thinking more on it, it is easier to explain that we just uniformly distribute everything over the 90% of the span. So I have changed it for that.

dataset_interfaces/interface.py
```diff
@@ -187,12 +189,51 @@ def load(cls, file_path: Path | str) -> "TestExample":
...
+        script_tokens = get_tokens_for_script(self.script)
+        assert script_tokens < memory_span, f"The script for test {self.dataset_name} is too long (estimated {script_tokens} tokens) for the specified memory span of {memory_span} tokens."
+
+        span_minus_script = memory_span - script_tokens
+
+        # Figure out what our spacing is going to be. Just distribute all the needles and questions across 90% of the memory span
+        token_wait = math.floor(span_minus_script * 0.9) // len(self.is_question)
```
dcasbol, line 212: The division should be by `N - 1`, since that's the number of waits that fit between N script lines.

dataset_interfaces/interface.py
```diff
@@ -187,12 +189,51 @@ def load(cls, file_path: Path | str) -> "TestExample":
...
+            if f != {}:
+                continue
+
+            self.waits[idx] = WaitCreator.create_wait(tokens=int(token_wait))
+            total_wait_tokens += token_wait
+
+        assert total_wait_tokens < span_minus_script, "Sum of task waits is higher than the determined memory span."
+        self.waits[-1] = WaitCreator.create_wait()
+
+    def setup_question_ratio(self):
```
dcasbol, line 231: Is this function used anywhere?

JosephDavidsonKSWH: Its not! Just a prototype that was left in.

PR #50, opened 2024-04-09
LTM-134 - Results aggregation

dcasbol:
Results from multiple examples of the same test are aggregated and a standard deviation is computed from them. This deviation is shown in the overall score of the detailed report.

The speed metric has been changed from `tests per hour`, which is too configuration dependent, to `tokens per second`, which is generic and can be computed now that we record agent tokens and duration separately.

**NOTE:** Testing it with old results will fail due to the absence of `agent_tokens` in the global statistics file.

### DIFF
--- a/reporting/templates/detailed_report.html
+++ b/reporting/templates/detailed_report.html
@@ -88,7 +88,7 @@
         <img src="data:image/png;base64, {{ logo_b64 }}" alt="GoodAI Logo" />
         <h1>GoodAI LTM Benchmark</h1>
         <p><b>Target Max Memory Span</b>: {{ target_memory_span }} tokens<br><b>Actual Test Memory Spans (Min/Avg/Max)</b>: {{ min_gap }} <b>/</b> {{ avg_gap }} <b>/</b> {{ max_gap }} tokens</p>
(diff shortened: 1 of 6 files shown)


reporting/templates/detailed_report.html
```diff
@@ -88,7 +88,7 @@
         <img src="data:image/png;base64, {{ logo_b64 }}" alt="GoodAI Logo" />
         <h1>GoodAI LTM Benchmark</h1>
         <p><b>Distraction Segment</b>: {{ info_gap }} tokens<br><b>Memory Span (Min/Avg/Max)</b>: {{ min_gap }} <b>/</b> {{ avg_gap }} <b>/</b> {{ max_gap }} tokens</p>
-        <p><b>Agent</b>: {{ agent_name }}<br><b>Overall score</b>: {{ achieved_score }} <b>/</b> {{ max_score }}</p>
+        <p><b>Agent</b>: {{ agent_name }}<br><b>Overall score</b>: {{ achieved_score }} <b>/</b> {{ max_score }} <b>±</b> {{ score_std }}</p>
```
JosephDavidsonKSWH, line 91: Can we not also display the STD for each dataset at the start of their own `<div>`?

JosephDavidsonKSWH (approved):

reporting/templates/detailed_report.html
dcasbol: yeah, I'll give it a try.

PR #51, opened 2024-04-10
LTM 138 -  LTM Claude

dcasbol:
Added support for LTM Agents with specified variant and core LLM.
`ltm_agent_<variant>[(<model>)]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,13 @@ langchain_kg_a    # Using 3.5-turbo-instruct and a knowledge graph memory
 langchain_ce_a    # Using 3.5-turbo-instruct and a conversation entity memory
 
 # GoodAI LTM models
-ltm_agent_1             # GoodAI LTM, 4-turbo preview, semantic retrieval + query generation + JSON scratchpad
-ltm_agent_2             # GoodAI LTM, 4-turbo preview, semantic retrieval
-ltm_agent_3             # GoodAI LTM, 4-turbo preview, semantic retrieval + text scratchpad
+# Variants:
+#   1. semantic retrieval + query generation + JSON scratchpad
+#   2. semantic retrieval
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -57,9 +57,13 @@ langchain_kg_a    # Using 3.5-turbo-instruct and a knowledge graph memory
...
 # GoodAI LTM models
-ltm_agent_1             # GoodAI LTM, 4-turbo preview, semantic retrieval + query generation + JSON scratchpad
-ltm_agent_2             # GoodAI LTM, 4-turbo preview, semantic retrieval
-ltm_agent_3             # GoodAI LTM, 4-turbo preview, semantic retrieval + text scratchpad
+# Variants:
+#   1. semantic retrieval + query generation + JSON scratchpad
+#   2. semantic retrieval
+#   3. semantic retrieval + text scratchpad
+# Optional model ID to use as core LLM
+# Example: ltm_agent_1(claude-3-opus-20240229)
```
JosephDavidsonKSWH, line 65: So in the case the default model would be `ltm_agent_1()`?

dcasbol: The brackets are entirely optional, so it would be `ltm_agent_1`. Does the alternative make more sense?

JosephDavidsonKSWH: On one hand, it might be more consistent with the parameterisation, on the other, that parameterisation is inconsistent with all the other models. Could you write the parsing to accept either `agent()` or `agent`?

dcasbol: It actually works already, because the empty string will be evaluated as `False` in the expression:
```python
model = params["model"] or GPT_4_TURBO_BEST
```
So yeah... both work!

JosephDavidsonKSWH: Great!

JosephDavidsonKSWH (approved):

PR #52, opened 2024-04-10
changed response to sneeze to something less guessable

JosephDavidsonKSWH:
The "Bless You!" response is too easy to guess correctly, so something sneeze related but more difficult to accidentally guess replaces it.


dcasbol (approved):
Nice touch

PR #56, opened 2024-04-12
Ltm 137 litellm

JosephDavidsonKSWH:
- Litellm performs the llm calls on the test side
- GPT and Claude models are now addressed by the same class `LLMChatSession`
- Litellm is used to count tokens based on model. Default model is Anthropics.
- Removed langchain agents, and `ResetPolicy` things
- `ChatSession`now supplies a `token_len()` function. By default its OpenAIs, due to the prevalence of the OpenAI stuff
- `LLMChatSession` uses the tokeniser based on the current LLM model.
- `utils.llm` has tools for counting tokens in scripts and lines for when we don't have access to an agent (like at test generation), by default it uses the anthropic tokeniser because it is less efficient
- GPT-4-0125 was broadly a mistake and the new `gpt-4-turbo` defaults to the new model


dataset_interfaces/interface.py
```diff
@@ -213,8 +213,8 @@ def default_waits(self):
         assert len(self.waits) == 0 or len(self.waits) == len(self.is_question), "Current waits should be empty or the same length as the script"
         memory_span = self.dataset_generator.memory_span
 
-        # TODO: All this token counting is based on OpenAIs methods, which is not Anthropics - but Aanthopic has no reliable tokeniser released so....
-        script_tokens = get_tokens_for_script(self.script)
+        # This is done at generation time, so we pick the least efficient encoder
```
dcasbol, line 216: I was confused by the wording. In this case, I guess that "less efficient" means that it uses more tokens? I'd put that instead, and why it is a better default.

model_interfaces/cost_estimation.py
```diff
@@ -1,6 +1,5 @@
 from model_interfaces.interface import ChatSession
 from dataclasses import dataclass, field
-from utils.text import token_len
```
dcasbol, line 3: Used in line 30:
```python
assert self.max_prompt_size > token_len(self.system_prompt)
```

model_interfaces/llm_interface.py
```diff
@@ -59,25 +101,11 @@ def _ts_message(m: dict) -> dict:
...
-            model=self.model,
-            temperature=0,
-            context_length=self.max_prompt_size,
-            cost_callback=cost_callback,
-        )
-        if self.verbose:
-            print(f"AGENT: {response}")
-        return response
+        self.context = self.build_context(user_message)
+        return super().reply(user_message)
```
dcasbol, line 105: `self.context` is always overwritten and `self.history` is never updated, so this will have no context at all in practice. Also, the user message will be added twice. What about wrapping the superclass' method?
```python
original_context = self.context
self.context = self.build_context()  # Without user message
response = super().reply(user_message)
self.context = original_context + self.context[-1:]
return response
```

dcasbol: I see the need for `self.history`, since you want to keep all the timestamps, but it also needs to be updated.

utils/filling_task.py
```diff
@@ -41,14 +38,14 @@ def filler_no_response_tokens_shakespeare(num_tokens: int, encoding_name="cl100k
     return filler_messages
 
 
-def filler_no_response_tokens_trivia(num_tokens: int, max_message_size: int):
+def filler_no_response_tokens_trivia(num_tokens: int, max_message_size: int, agent: ChatSession):
```
dcasbol, line 41: It might be better to pass just the token counting function, but no drama.

utils/llm.py
```diff
@@ -2,37 +2,36 @@
...
 MODEL_ALIASES = {
     "gpt-3.5-turbo": "gpt-3.5-turbo-0125",
-    "gpt-4-turbo": "gpt-4-0125-preview",
+    "gpt-4-turbo": "gpt-4-turbo-2024-04-09",
     "gpt-4-1106": "gpt-4-1106-preview",
-    "gpt-4-0125": "gpt-4-0125-preview",
     "claude-3-haiku": "claude-3-haiku-20240229",
     "claude-3-sonnet": "claude-3-sonnet-20240229",
     "claude-3-opus": "claude-3-opus-20240229",
 }
```
dcasbol, line 32: Can we maybe get rid of all this and use litellm's logic?

utils/llm.py
```diff
@@ -64,7 +63,7 @@ def token_cost(model: str) -> tuple[float, float]:
     return cost_info
 
 
-def response_cost(response: openai.ChatCompletion) -> float:
+def response_cost(response: ModelResponse) -> float:
```
dcasbol, line 66: `litellm` has [several functions](https://docs.litellm.ai/docs/completion/token_usage) that could replace ours, like `completion_cost`.

utils/llm.py
```diff
@@ -81,26 +80,20 @@ def ensure_context_len(
...
-        message_tokens = context_token_len([message], model=model)
+    for message in reversed(context[1:]):
+        message_tokens = litellm.token_counter(model, text=message["content"])
         if context_tokens + message_tokens + response_len > max_len:
             break
         messages.append(message)
         context_tokens += message_tokens
     messages.reverse()
-    context = context[:sys_idx] + messages
+    context = context[:1] + messages
```
dcasbol, line 96: Maybe `litellm` performs auto-trimming too?

utils/llm.py
```diff
@@ -112,18 +105,14 @@ def ask_llm(
     context_length: int = None,
     cost_callback: Callable[[float], None] = None,
     timeout: float = 300,
-    max_tokens: Optional[int] = None,
+    max_response_tokens: int = 1000,
```
dcasbol, line 108: Is there any reasoning behind this default value? `None` means that it will be allowed to generate until it fills up the context or whatever the API says.

utils/llm.py
```diff
@@ -112,18 +105,14 @@ def ask_llm(
...
-    context, context_tokens = ensure_context_len(context, model, context_length, response_len=256)
-    response = openai.chat.completions.create(
-        model=model,
-        messages=context,
-        temperature=temperature,
-        timeout=timeout,
-        max_tokens=max_tokens,
-    )
+    context, context_tokens = ensure_context_len(context, model, context_length, response_len=max_response_tokens)
+    response = completion(model=model, messages=context, max_tokens=max_response_tokens, temperature=temperature)
```
dcasbol, line 114: missing `timeout`

utils/llm.py
```diff
@@ -147,23 +136,26 @@ def make_assistant_message(content: str) -> LLMMessage:
 
 
 def context_token_len(context: LLMContext, model: Optional[str] = None, response: bool = False) -> int:
```
dcasbol, line 138: It looks like [`litellm.token_counter`](https://litellm.vercel.app/docs/completion/token_usage#3-token_counter) takes care of this.

dcasbol:
In the description you mention: "Litellm is used to count tokens based on model. **Default model is Anthropics**." But I can't see where that is in the code.

utils/llm.py
JosephDavidsonKSWH: Oh I didn't know that one. I thought I needed some ints there.

utils/llm.py
```diff
@@ -2,130 +2,98 @@
...
-    "gpt-4-0125": "gpt-4-0125-preview",
     "claude-3-haiku": "claude-3-haiku-20240229",
     "claude-3-sonnet": "claude-3-sonnet-20240229",
     "claude-3-opus": "claude-3-opus-20240229",
 }
 GPT_CHEAPEST = "gpt-3.5-turbo"
-GPT_4_TURBO_BEST = "gpt-4-1106-preview"
-client = OpenAI()
+GPT_4_TURBO_BEST = "gpt-4-turbo"
+CHEAPEST_TOKENISER = "claude-3-opus"
```
dcasbol, line 24: why cheapest? if it's because it gives a higher token count, I'd use a different name that expresses that.

utils/llm.py
```diff
@@ -2,130 +2,98 @@
...
+litellm.model_alias_map = MODEL_ALIASES
```
dcasbol, line 25: I thought that `litellm.model_alias_map` was filled automatically with the standard aliases. Is this overriding it? Am I wrong and it's empty by default?

utils/llm.py
```diff
@@ -146,24 +114,21 @@ def make_assistant_message(content: str) -> LLMMessage:
...
 
-def get_tokens_for_script(script: list[str], model: Optional[str] = None) -> int:
-    encoder = tiktoken.encoding_for_model(get_model(model))
+def tokens_in_script(script: list[str], model: str = CHEAPEST_TOKENISER) -> int:
+    model = model_from_alias(model)
     num_tokens = 0
     for line in script:
         num_tokens += 4
-        num_tokens += len(encoder.encode(line))
+        num_tokens += litellm.token_counter(model, text=line)
```
dcasbol, line 127: The `+= 4` is inherited from the old `context_token_len`, which was intended for use with OpenAI models (taken [from here](https://github.com/openai/openai-cookbook/blob/main/examples/)). I'd add a comment or something, so that we don't forget why that number is there e.g. `# every message follows <|start|>{role/name}\n{content}<|end|>\n`

model_interfaces/interface.py
```diff
@@ -70,7 +62,6 @@ def save(self):
     def load(self):
         pass
 
-
-class ChatSessionFactory(ABC):
-    def create_session(self) -> ChatSession:
-        pass
+    def token_len(self, text: str) -> int:
+        enc = tiktoken.get_encoding("cl100k_base")
+        return len(enc.encode(text))
```
dcasbol, line 67: Why is this class taking a different route for computing the number of tokens?

utils/llm.py
JosephDavidsonKSWH: It is empty by default.

utils/llm.py
```diff
@@ -2,130 +2,98 @@
...
 def ask_llm(
     context: LLMContext,
-    model: Optional[str] = None,
+    model: str,
     temperature: float = 1,
     context_length: int = None,
     cost_callback: Callable[[float], None] = None,
     timeout: float = 300,
-    max_tokens: Optional[int] = None,
+    max_response_tokens: Optional[int] = None,
```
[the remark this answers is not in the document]
JosephDavidsonKSWH: Not as it currently stands, because `ensure_context_len` uses it and expects it to be an integer. What I can do is just get the `max_tokens` for the model right after getting the model name.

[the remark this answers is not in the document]
dcasbol: Ah I understand now. Yeah, you can do that and also use the typing
```python
max_response_tokens: int = None
```
which should make clear that the value is expected to be filled at some point, and the null value is just there to ease checks.

model_interfaces/interface.py
JosephDavidsonKSWH: So this is here as the default tokeniser for a model, since not all models should supply their own. Since GPT is more common, I decided that should be the default. Though I could now just have this function call the `utils.llm` functions for tokenising with a GPT model, and comment why its doing that.

dcasbol: Yeah, I don't know what's best. I could agree with any of them "most common" or "out of caution", but the comment is a good thing I think.

dcasbol (approved):

PR #57, opened 2024-04-12
Ltm 145 cheaper llms

JosephDavidsonKSWH:
Note: This is diffed from the litellm branch as the changes here are dependent on it.

- Adding of `agent_response` to `send_message`
- `LLMChatSession` makes use of the response.
- The response is filled if `SendMessageAction.is_filling == True` and there is a value for `filler_response`.
- Costs are a little borked

### DIFF
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -61,6 +61,7 @@ class SendMessageAction(TestAction):
     sent_ts: Optional[datetime] = None
     is_question: bool = False
     is_filling: bool = False
+    filler_response: str = None
 
 
 @dataclass
--- a/model_interfaces/interface.py
+++ b/model_interfaces/interface.py
@@ -15,10 +15,10 @@ class ChatSession(ABC):
     is_local: bool = False
     max_message_size: int = 1000
 
-    def message_to_agent(self, user_message: str) -> Tuple[str, datetime, datetime]:
+    def message_to_agent(self, user_message: str, agent_response: str = "") -> Tuple[str, datetime, datetime]:
         sent_ts = datetime.now()
         old_costs = self.costs_usd
-        response = self.reply(user_message)
+        response = self.reply(user_message, agent_response)
         reply_ts = datetime.now()
         assert (
             self.is_local or old_costs < self.costs_usd
@@ -41,7 +41,7 @@ class ChatSession(ABC):
         return f"{self.run_name} - {self.name}"
 
     @abstractmethod
-    def reply(self, user_message: str) -> str:
+    def reply(self, user_message: str, agent_response: str) -> str:
         """
         In this method, the agent is expected to:
         - Generate a response to "user_message" and return it as a plain string.
--- a/model_interfaces/interface.py
+++ b/model_interfaces/interface.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Tuple
+from typing import Tuple, Optional
 from dataclasses import dataclass
 
 import tiktoken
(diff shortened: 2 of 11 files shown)


model_interfaces/interface.py
```diff
@@ -41,7 +41,7 @@ def save_name(self):
         return f"{self.run_name} - {self.name}"
 
     @abstractmethod
-    def reply(self, user_message: str) -> str:
+    def reply(self, user_message: str, agent_response: str) -> str:
```
dcasbol, line 44: It's a bit confusing that `agent_response` is mandatory. I'd maybe make it optional with a default value of `None`. Also don't forget that the empty string is also a valid message for an agent to give.

model_interfaces/llm_interface.py
```diff
@@ -31,16 +31,20 @@ def __post_init__(self):
...
 
         self.context.append(make_user_message(user_message))
-        response = ask_llm(self.context, self.model, context_length=self.max_prompt_size, cost_callback=cost_callback,
-                           max_response_tokens=self.max_response_tokens)
+        if agent_response == "":
+            response = ask_llm(self.context, self.model, context_length=self.max_prompt_size, cost_callback=cost_callback,
+                               max_response_tokens=self.max_response_tokens)
+        else:
+            response = agent_response
+            self.costs_usd += 1e-6
```
dcasbol, line 47: You could maybe skip the assertion if `agent_response is not None`.

dcasbol (approved):
I think it's good to merge, apart from the commented lines.

model_interfaces/interface.py
JosephDavidsonKSWH: We should figure this out. Reply is only called by `ChatSession.message_to_agent()` which has the optional `agent_response`, that that optional parameter is given to `reply`. Should the optional param in message_to_agent, also be optional for reply?.

dcasbol: I don't think it needs to be optional, but the default value should be `None`. `None` means "ignore this value", and otherwise it should be taken as a response. But making it optional would feel better, sure.

JosephDavidsonKSWH: In my mind, all values that have defaults are optional. Since they are optional at call time.

dcasbol: Ah right. Sorry. I expressed it badly. I meant that `None` should be given by default by `message_to_agent`, and then I think it's better if the parameter is optional in `reply`, because it's more flexible, but there's nothing forcing it to be right now.

model_interfaces/interface.py
```diff
@@ -15,14 +15,16 @@ class ChatSession(ABC):
...
+        response = self.reply(user_message, agent_response=agent_response)
         reply_ts = datetime.now()
-        assert (
-            self.is_local or old_costs < self.costs_usd
-        ), "The agent implementation is not providing any cost information."
+        # If we are supplying a response from the agent, then don't count costs.
+        if agent_response is not None:
+            assert (
+                self.is_local or old_costs < self.costs_usd
+            ), "The agent implementation is not providing any cost information."
```
dcasbol, line 27: It's the other way round 🙃 `if agent_response is None:` to run the assert when the agent actually generates an answer.

JosephDavidsonKSWH: Yeah my testing discovered that :P

dataset_interfaces/interface.py
```diff
@@ -61,6 +61,7 @@ class SendMessageAction(TestAction):
     sent_ts: Optional[datetime] = None
     is_question: bool = False
     is_filling: bool = False
+    filler_response: str = None
```
dcasbol, line 64: This is actually optional, isn't it?

PR #58, opened 2024-04-15
Ltm 144 waits after reply

JosephDavidsonKSWH:
Claude tests seem to overrun because Claude is very verbose in its messages. One potential fix is in taking the reply to previous statement before the wait into account when calculating how long to wait for.

### DIFF
--- a/dataset_interfaces/interface.py
+++ b/dataset_interfaces/interface.py
@@ -210,36 +210,29 @@ class TestExample:
         if isinstance(self, DynamicExample):
             return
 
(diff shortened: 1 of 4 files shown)


dcasbol:
I guess that by "when calculating how long to wait for" you mean at generation time, right? But that would benefit other models. Maybe the percentages is the answer to this? Instead of fixed-token waits, the percentages would make the script lines land more or less where they should be.

dcasbol (approved):

PR #60, opened 2024-04-16
Ltm 147 prospective

JosephDavidsonKSWH:
Prospective memory had issues where the task was not understood by the LLMs.

- Removed the GPT generation part of the dataset - examples are now chosen from real quotes.
- Question phrasing is changed to Finalys original. Which the LLMs seem to understand best. 
- Callback ends at the first moment the test is passed or failed

### DIFF
--- a/datasets/prospective_memory.py
+++ b/datasets/prospective_memory.py
@@ -37,6 +37,7 @@ def cites_quote(quote: str, message: str) -> bool:
 class ProspectiveMemoryDataset(DatasetInterface):
     name: str = "Prospective Memory"
     description: str = "Give the agent a fictitious quote, then ask it to append to the nth reply"
+    reset_message: str = "Forget my instruction to append a quote to one of your replies."
     # generation_file: Path = DATA_DIR.joinpath("gpt_generation_prompts/2-1_prospective_memory_test.json")
     # temperature: float = 0.5
     # uses_callback: bool = True
--- a/datasets/prospective_memory.py
+++ b/datasets/prospective_memory.py
@@ -1,4 +1,3 @@
-import re
 import string
 from dataclasses import dataclass
 from typing import List, Tuple, Any, Iterator
@@ -93,7 +92,7 @@ class ProspectiveMemoryDataset(DatasetInterface):
                 break
 
         agent_responses = task_log[statement_idx + 1:][::2]
-        response_w_quote_idx = int(re.findall(r"\d+", question)[0]) - 1
+        response_w_quote_idx = example.expected_responses[0][2] - 1
 
         max_score = 1
 
--- a/datasets/prospective_memory.py
+++ b/datasets/prospective_memory.py
@@ -53,16 +53,16 @@ class ProspectiveMemoryDataset(DatasetInterface):
         for _, request_num in zip(range(num_examples), self.request_num_generator()):
             script = []
             # Pick a quote
-            quote_pair = QUOTES[request_num]
+            quote, author = QUOTES[request_num]
             target = self.random.randint(2, 8)
 
             n = ordinal(target)
 
-            script.append(f"'{quote_pair[0]}' - {quote_pair[1]}")
-            script.append(f'After responding to the following unrelated user prompts, please append the quote from {quote_pair[1]}, to your {n} response. Count your response to this message as the first response.')
+            script.append(f"'{quote}' - {author}")
+            script.append(f'After responding to the following unrelated user prompts, please append the quote from {quote}, to your {n} response. Count your response to this message as the first response.')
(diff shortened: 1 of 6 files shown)


datasets/prospective_memory.py
```diff
@@ -19,22 +28,51 @@ def cites_quote(quote: str, message: str) -> bool:
...
-            n = ordinal(self.random.randint(2, 9))
-            example.script[1] = re.sub(num_pattern, n, example.script[1])
-            example.script[1] += " Count your response to this message as the first response."
-            example.expected_responses[0] = re.sub(num_pattern, n, example.expected_responses[0])
+        examples = []
+
+        for _, request_num in zip(range(num_examples), self.request_num_generator()):
+            script = []
+            # Pick a quote
+            quote_pair = QUOTES[request_num]
```
dcasbol, line 56: Mainly being picky here, but you could do
```python
quote, author = QUOTES[request_num]
```
and then
```python
script.append(f"{repr(quote)} - {author}")
```

datasets/prospective_memory.py
```diff
@@ -54,7 +92,7 @@ def continual_evaluation_callback(
                 break
 
         agent_responses = task_log[statement_idx + 1:][::2]
-        response_w_quote_idx = int(re.findall(r"\d+", question)[0]) - 1
+        response_w_quote_idx = example.expected_responses[0][2] - 1
```
dcasbol, line 95: Suggestion:
```python
quote, author, nth = example.expected_responses[0]
```

datasets/prospective_memory.py
```diff
@@ -68,22 +118,14 @@ def continual_evaluation_callback(
...
-        quote = None
-        for quote_chars in ["''", '""', "``", "<>"]:
-            open_char, close_char = quote_chars
-            pattern = re.compile(f"{re.escape(open_char)}(.*?){re.escape(close_char)}")
-            matches = pattern.findall(example.script[0])
-            if len(matches) > 0:
-                quote = matches[0].lower()
-                break
-        assert quote is not None, "Quote cannot be found"
+        quote = example.expected_responses[0][0]
```
dcasbol, line 121: redundant

runner/master_log.py
```diff
@@ -156,6 +168,20 @@ def messages(self, test_id: str = "") -> list[str]:
...
+        for event in self.log:
+            if event.type in [EventType.SEND_MESSAGE, EventType.RESPONSE_MESSAGE, EventType.SEND_FILL, EventType.RESPONSE_FILL]:
+                if test_id == event.test_id or getting_all_messages:
+                    sender = "Test" if event.type == EventType.SEND_MESSAGE else "System" if event.type == EventType.SEND_FILL else "Agent"
+                    messages.append(f"{sender} ({event.timestamp}): {event.data['message']}")
+
+                    if event.data["is_question"]:
+                        getting_all_messages = True
+
+        return messages
```
dcasbol, line 183: I have trouble understanding this method. Why is `getting_all_messages` set only when it finds a question? It's starting from the very first logged event, so it could be any question.

dcasbol: Ah I see. The question is when the callback is registered. Alright 👌🏾

datasets/prospective_memory.py
```diff
@@ -19,29 +28,69 @@ def cites_quote(quote: str, message: str) -> bool:
...
+        first_message = example.script[0]
+        start_idx = 0
+        all_messages = scheduler.master_log.messages()
+        for idx, msg in enumerate(all_messages):
+            if first_message in msg:
+                start_idx = idx
+                break
+
+        return all_messages[start_idx:]
+
```
dcasbol, line 93: Not used.

datasets/prospective_memory.py
```diff
@@ -19,29 +28,69 @@ def cites_quote(quote: str, message: str) -> bool:
...
+        for _, request_num in zip(range(num_examples), self.request_num_generator()):
+            script = []
+            # Pick a quote
+            quote, author = QUOTES[request_num]
+            target = self.random.randint(2, 8)
+
+            n = ordinal(target)
+
+            script.append(f"'{quote}' - {author}")
```
dcasbol, line 61: Suggestion: `repr(quote)` will add the quotation marks, and make them single or double depending on the content of the quote.

runner/master_log.py
dcasbol: This method is almost the same as `master_log.messages(test_id)`. Could it work to rename it to `master_log.test_readable_log(test_id, is_callable)`?

dcasbol (approved):
Comments are mainly suggestions.

PR #70, opened 2024-04-19
LTM-143 - Accurate progress bar

dcasbol:
The bar is updated with every message sent. The progress is very accurately estimated by means of the token count and the expected test memory span. In the text field, the current average score and standard deviation is shown.

### DIFF
--- a/reporting/results.py
+++ b/reporting/results.py
@@ -5,6 +5,7 @@ import json
 from utils.files import make_result_path, parse_result_path
 from dataset_interfaces.factory import DATASETS_BY_NAME
 
+
 @dataclass
 class TestResult:
     run_name: str
@@ -54,6 +55,10 @@ class TestResult:
         string += f"Tokens: {self.tokens}\n"
         string += f"Characters: {self.characters}\n"
         return string
+
+    @property
+    def unique_id(self):
+        return f"{self.dataset_name} - {self.example_id}"
 
     @property
     def path(self) -> Path:
--- a/runner/progress.py
+++ b/runner/progress.py
@@ -2,13 +2,25 @@ import tkinter as tk
 from tkinter import ttk
 from dataset_interfaces.interface import TestExample
 from reporting.results import TestResult
+from collections import defaultdict
+from utils.math import mean_std
+
+
+def blinker_gen():
+    while True:
+        for c in r"/-\|":
+            yield c
 
 
 class ProgressDialog(tk.Tk):
     def __init__(self, num_tests: int):
         super().__init__()
         self._num_tests = num_tests
-        self._tests_run = 0
+        self._memory_span = None
+        self._at = 0
+        self._test_info = dict()
+        self._scores = defaultdict(lambda: list())
+        self._blinker = blinker_gen()
 
         self.title("GoodAI LTM Benchmark")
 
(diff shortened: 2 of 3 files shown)


runner/scheduler.py
```diff
@@ -263,7 +264,6 @@ def fast_forward_tests(self, tests: dict[str, TestExample]):
             match evt.type:
                 case EventType.BEGIN:
                     result, skip = self.initialise_result(test)
-                    assert not skip
```
dcasbol, line 266: I added this "fix". It did fix resuming on the run with Claude, but I'm not 100% sure about it.

reporting/results.py
```diff
@@ -55,6 +56,10 @@ def __str__(self):
         string += f"Characters: {self.characters}\n"
         return string
 
+    @property
+    def unique_id(self):
+        return f"{self.dataset_name} - {self.example_id}"
```
JosephDavidsonKSWH, line 61: Isn't this nomenclature confusing? Pretty sure I see `unique_id == example_id` in the scheduler.

runner/scheduler.py
JosephDavidsonKSWH: Has it ever failed "legitimately"?

reporting/results.py
dcasbol: It is. And that is why I added a card in JIRA. Currently:
- `example_id`: index of the sample within the dataset (assigned during generation, internal to the dataset).
- `unique_id`: an ID that is unique across the system.

There's also `TestExample.repetition`, which we actually don't use, but it's displayed in the result filename.

runner/scheduler.py
dcasbol: Maybe? I don't know. Early fails were due to system failures, so many things were wrong. Now the system is more stable and it hasn't even failed so often during resuming.

runner/progress.py
```diff
@@ -20,14 +32,37 @@ def __init__(self, num_tests: int):
         self.update_idletasks()
 
     def notify_running(self, example: TestExample):
-        self._label.config(text=f"Running {example.dataset_name} ({example.example_id})")
-        self.update_idletasks()
+        self._at = max(self._at, example.start_token)
+        self._memory_span = self._memory_span or example.dataset_generator.memory_span
+        self._test_info[example.unique_id] = dict(start=example.start_token, span=self._memory_span)
```
JosephDavidsonKSWH, line 37: If `_memory_span` is an attribute of the class, its going to be same for all elements of `_test_info`

dcasbol: Yes, but it's updated when the test actually finishes, which might be earlier than that. It's probably not worth the extra thought, but it helped me figure it out.

dcasbol: (Maybe I wasn't clear enough) `info["span"]` is updated when the test finishes, so that it actually shows a 100% complete when the progress bar is recomputed. Otherwise it would get stuck at 90% or so.

JosephDavidsonKSWH: You mean at line 46? I'm not so sure... Because its updated to be the max value of the current token, and when it starts, vs the span. Since all tests come in under the memory span, `info["span"]` is just the memory span value again.

dcasbol: Ah right! I should have taken the minimum of both. Noted! 👌🏾

dcasbol: In the end I just took the actual span. After all, it will be the more accurate regardless it being shorter or longer.

JosephDavidsonKSWH (approved):

PR #71, opened 2024-04-23
Restaurant edge case fix

dcasbol:


datasets/restaurant.py
```diff
@@ -128,16 +128,18 @@ def extract_order_items(self, message: str) -> list[str]:
...
+        for item in sh_order:  # Sometimes there is a drink in the order, which is not in the menu.
+            for section_content in self.dataset_generator.menu_dict.values():
+                for section_item in section_content:
+                    if item in section_item:
+                        choices = [c for c in section_content if c not in [section_item, old_item]]
+                        new_item = self.random.choice(choices)
+                        altered_order = [item for item in order]
+                        i = altered_order.index(item)
+                        altered_order[i] = new_item
+                        return section_item, new_item, altered_order
```
JosephDavidsonKSWH, line 142: So If I am reading this correctly, you shuffle the parsed order, then basically pick the first item (assuming all parsed items are not hallucinations) and pick something new from that section to be the mistaken order?

dcasbol: Yeah, the shuffling is just to avoid picking always the first item (i.e. alternative to the `choice` that was previously in place), but still have the option to take another one if the chosen item happens to be a drink (drinks are given more freedom and not shown exactly as they are in the menu).

PR #72, opened 2024-04-23
Benchmark v3

JosephDavidsonKSWH:


README.md
```diff
@@ -86,15 +86,8 @@ The datasets that are implemented for this benchmark can be found in `./datasets
...
 colours
-conflicting_personal_info
-delayed_recall
-how_to_think
-instruction_recall
 jokes
-kv
-locations
 locations_directions
-names
```
dcasbol, line 97: These datasets might not be 100% reliable and therefore not included in the final configuration, but they are still implemented in the repository.

README.md
```diff
@@ -118,33 +111,45 @@ The repository consists of four parts:
 More details for each of these parts can be found here: [datasets](datasets/README.md), [models](model_interfaces/README.md), [runner](runner/README.md), [reports](reporting/README.md).
 
 
-## Benchmark 2 - 03/2024
+## Benchmark 3 - 04/2024
```
dcasbol, line 114: The section "Configurations" seems outdated. Also, what do you think about placing the old configuration files under `configurations/older` or something like that?

model_interfaces/joe_ltm_prototype.py
dcasbol, line 1: Do you want to keep this empty file?

JosephDavidsonKSWH: No, it is getting removed.

README.md
JosephDavidsonKSWH: I guess... I can restore it.

PR #78, opened 2024-05-01
Show results by metric

dcasbol:
One figure is generated for every metric. There is still the possibility to filter out some agents if we want to show only pure LLMs or rather LTM agents, since it's hard to interpret the figure if there are too many values on it. In order to do that, just uncomment and maybe change lines `56-57`.

```python
for name, short in agent_names:
    #if not name.startswith("LLMChatSession"):
    #    continue
```


JosephDavidsonKSWH (approved):

JosephDavidsonKSWH:
I think the LTM score should be reworked in another PR

PR #80, opened 2024-05-15
Re-compute speed metrics from master log

dcasbol:


dev_scripts/figures.py
```diff
@@ -1,6 +1,43 @@
...
+            t0 = evt.timestamp
+        else:
+            msg = evt.data["message"]
+            t = (evt.timestamp - t0).total_seconds()
+            tokens = message_tokens(msg, agt_name)
+            speeds.append(tokens / t)
+            total_tokens += tokens
+            total_time += t
+    # return mean_std(speeds)
+    return total_tokens / total_time, 0
```
JosephDavidsonKSWH, line 40: So we are counting response times for the test messages only?

JosephDavidsonKSWH (approved):

dev_scripts/figures.py
dcasbol: Ah yes... that was me trying to get a reliable speed measurement. LLMs don't actually reply to filler messages, so this is how I managed to skip them. I could still consider the fillers for LTM agents though. However, I am still not happy with the results... I will try a couple of things more today and send you some previews.

PR #81, opened 2024-05-17
Hf hacks

JosephDavidsonKSWH:


utils/llm.py
```diff
@@ -31,7 +31,10 @@ def model_from_alias(model: str):
 
 def get_max_prompt_size(model: str):
     model = model_from_alias(model)
-    return litellm.model_cost[model]["max_input_tokens"]
+    model_info = litellm.model_cost.get(model, None)
+    if model_info:
+        return model_info["max_input_tokens"]
+    return 0
```
dcasbol, line 37: I'm not sure about this `return 0`. Maybe an error would be better?

utils/llm.py
```diff
@@ -120,7 +123,10 @@ def ask_llm(
         claude_adjust_factor += 0.2 * (actual_count / context_tokens)
         
     if cost_callback is not None:
-        cost_callback(litellm.completion_cost(response))
+        try:
+            cost_callback(litellm.completion_cost(response))
+        except:
+            cost_callback(0.001)
```
dcasbol, line 129: What's the reason of this fallback cost?

utils/llm.py
JosephDavidsonKSWH: It is to stop the error! Litellm doesn't have all possible models in its repository. So it breaks when trying to retireve tokens. In this case, we should just trust the user.

utils/llm.py
JosephDavidsonKSWH: Same issue with the litellm. But maybe we dont need it and can just set the model to be local instead?

dcasbol: Alright, understood. In the future we could actually prompt the user to add cost information and store it somewhere.

dcasbol (approved):

dcasbol (approved):

PR #92, opened 2024-05-30
Updated evaluations for restaurant and spy meeting

JosephDavidsonKSWH:
* Fix for restaurant task, where some LLMs say "Thank you for the <drink>" which is interpreted as an order
* Give partial credit for Spy Meeting, and a penalty for agent confusion.
* Updated evaluation script for new `TestExample` types and flow

### DIFF
--- a/datasets/restaurant.py
+++ b/datasets/restaurant.py
@@ -42,7 +42,7 @@ class RestaurantExample(DynamicExample):
         )
         self.expected_responses.append("The agent follows the role of a customer at a restaurant and orders a drink.")
         self.check_role_following()
-        drinks = self.extract_order_items(self.action.reply)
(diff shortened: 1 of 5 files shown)


datasets/restaurant.py
```diff
@@ -107,7 +107,7 @@ def extract_order_items(self, message: str) -> list[str]:
 
         items = list()
         for item_dict in response["order"]:
-            if item_dict["is_drink"]:
+            if item_dict["is_drink"] and should_order_drink:
```
dcasbol, line 110: (reminder) It will probably be better to have separate methods for drinks and food. This change results in the verbatim menu entry being added to the list (see line 116).

dcasbol (changes requested):
The restaurant task can be addressed later.

dcasbol (approved):

PR #105, opened 2024-06-07
Heat-map tables for supplementary material

dcasbol:


dev_scripts/heat_tables.py
```diff
@@ -0,0 +1,134 @@
...
+def set_alt_name(idx: int, mem_spans: list[str], alt_name: str):
+    ALIASES[alt_name] = ALIASES[TO_RETRIEVE["Benchmark 3 - 1k"][idx]]
+    for span in mem_spans:
+        TO_RETRIEVE[f"Benchmark 3 - {span}"][idx] = alt_name
+
+
+set_alt_name(0, ["1k", "32k"], "LLMChatSession - together_ai-mistralai-Mixtral-8x7B-Instruct-v0.1 - 32000")
+set_alt_name(4, ["120k", "200k"], "LLMChatSession - gpt-4-turbo - 128000")
+set_alt_name(10, ["120k", "200k", "500k"], "LTMAgentWrapper - claude-3-opus-20240229 - 16384 - QG_JSON_USER_INFO")
+set_alt_name(9, ["200k", "500k"], "LTMAgentWrapper - gpt-4-turbo-2024-04-09 - 16384 - QG_JSON_USER_INFO")
```
JosephDavidsonKSWH, line 58: I have no idea what these function calls are doing

JosephDavidsonKSWH: Ohhh, I think I see. The context sizes are a little different for these agents in some runs, so this is replacing those names under those runs.

JosephDavidsonKSWH (approved):

dev_scripts/heat_tables.py
dcasbol: With those calls I'm basically saying _"The agent at index 0 will take this other name for spans 1k and 32k"_. I needed it because the results were sometimes saved like `claude-3-opus` and other times as `claude-3-opus-20240229`.

PR #106, opened 2024-06-10
Add Colour Evaluator feature

FKGSOFTWARE:
This pull request adds the Colour Evaluator feature, which includes changes to colours.py, test_cases_colours.json, and test_colours.py. 

The feature enhances the color evaluation functionality through use of Regex, as well as providing some tests and test cases.


datasets/colours.py
```diff
@@ -31,10 +31,21 @@
 ]
 
 
+def match_color(text, colors):
+    """
+    Check if any of the exact words in 'colors' is in the text.
+    Returns 1 if any color is found as a whole word, otherwise 0.
+    """
+    colors_pattern = '|'.join([re.escape(color) for color in colors])
+    pattern = rf'(?<!\S)(?:{colors_pattern})(?!\S)|(?<!\S)(?:{colors_pattern})(?=[,.!?])'
```
dcasbol, line 40: Is this `(?!\S)` an optional white space? I think you can write it just `\s?`, but I'm not sure.

datasets/colours.py
```diff
@@ -31,10 +31,21 @@
...
+    match = re.search(pattern, text, re.IGNORECASE)
+    return 1 if match else 0
```
dcasbol, line 42: Why is it not returning a `bool`?

datasets/colours.py
```diff
@@ -71,10 +82,21 @@ def generate_examples(self, num_examples):
...
     def evaluate_correct(
-        self, questions: List[str], responses: List[str], expected_answers: List[str]
+            self, questions: List[str], responses: List[str], expected_answers: List[str]
     ) -> Tuple[int, int, List[str]]:
-        color = expected_answers[0].lower()
-        if color in responses[-1].lower():
-            return 1, 1, [f'"{color}" is in the response.']
-        return 0, 1, [f'"{color}" is NOT in the response.']
+        score = 0
+        max_score = len(expected_answers)
```
dcasbol, line 88: `max_score` will be deprecated at some point, but for now we should set it always to `1`.

datasets/test/test_cases_colours.json
```diff
@@ -0,0 +1,402 @@
...
+  },
+  {
+    "questions": ["What color is associated with the Green Giant?"],
+    "responses": ["The Green Giant is a brand of vegetables associated with the color green."],
+    "expected_answers": ["Green"]
+  },
+  {
+    "questions": ["What color do you love?"],
+    "responses": ["I love the color indigo, but it is not one of the listed colors."],
+    "expected_answers": [""]
```
dcasbol, line 395: When would we have an empty expected answer?

datasets/test/test_colours.py
```diff
@@ -0,0 +1,18 @@
...
+    dataset = ColourDataset()
+
+    for test_case in test_cases:
+        questions = test_case['questions']
+        responses = test_case['responses']
+        expected_answers = test_case['expected_answers']
+
+        score, max_score, response = dataset.evaluate_correct(questions, responses, expected_answers)
+
+        assert score == max_score, f"Test Case: {test_case}\nExpected: {max_score}\nResult: {score}\nResponse: {response}"
```
dcasbol, line 18: Because `test_case` shares names with the method's parameters, you could call it like `evaluate_correct(**test_case)` (just saying, it's good as you have it, and even clearer)

Apart from that, I guess from looking at the code that all tests in the JSON file are positive, right?

dcasbol:
I like the thoroughness of the test 👌🏾

datasets/colours.py
FKGSOFTWARE: The rationale was the pattern (?<!\S) is a negative lookbehind for a non-whitespace character, which is more precise than \s? in this context. Although I did stop when tests were passing for the element so could be improved.

datasets/colours.py
FKGSOFTWARE: Originally, the result was used as a conditional but is no longer so can probably be cleaned

datasets/colours.py
dcasbol: I don't understand the difference 😅 But as you said, the important part is that it passes the tests. An important detail would then be to have negative examples too, so that we know that the evaluation is not too loose.

datasets/test/test_colours.py
FKGSOFTWARE: Noted, I'll address that. 
Correct, all tests pass. There is one set of edge cases that I think I missed so I will investigate when I do the revisions.

datasets/test/test_cases_colours.json
FKGSOFTWARE: In these cases, "" denotes no colours are found of an appropriate form.

datasets/colours.py
FKGSOFTWARE: Will revise accordingly

datasets/colours.py
```diff
@@ -71,10 +83,21 @@ def generate_examples(self, num_examples):
...
+        for expected_color, response_text in zip(expected_answers, responses):
+            result = match_color(response_text, [expected_color])
+            if expected_color == "" and result is False:
+                score += 1
+                response_messages.append("No color expected in the response.")
+            elif result:
+                score += 1
+                response_messages.append(f'"{expected_color}" is in the response.')
+            else:
+                response_messages.append(f'"{expected_color}" is NOT in the response.')
```
dcasbol, line 101: The scoring function should take care that the score is not greater than one. You can probably just divide it by the number of expected responses.

dcasbol (approved):

PR #111, opened 2024-06-26
Loading from state now sets local attributes in object instead of creating a new object

JosephDavidsonKSWH:
…

### DIFF
--- a/ltm/agent.py
+++ b/ltm/agent.py
@@ -108,10 +108,7 @@ class LTMAgent:
         )
         return json.dumps(state, cls=SimpleJSONEncoder)
 
-    @classmethod
-    def from_state_text(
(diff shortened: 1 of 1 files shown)


dcasbol:
Wasn't it an easier solution (not sure if better) to change [line 45 in the wrapper](https://github.com/GoodAI/goodai-ltm-benchmark/blob/dev/model_interfaces/ltm_agent_wrapper.py#L45) for this one?:
```python
self.agent = LTMAgent.from_state_text(fd.read())
```

JosephDavidsonKSWH:
I did consider that. But at this point in the execution, there is already an `LTMAgent` object. I think both ways are equivalent, but using the already existing object made more sense to me.

dcasbol (approved):

PR #119, opened 2024-07-09
ask_llm does not modify the context but performs checks

dcasbol:


utils/llm.py
```diff
@@ -67,81 +47,68 @@ def ensure_context_len(
...
     model = model_from_alias(model)
     max_len = max_len or get_max_prompt_size(model)
-    messages = list()
-    if len(context) > 0 and context[0]["role"] == "system":
+    if context[0]["role"] == "system":
         sys_prompt = context[:1]
         context = context[1:]
     else:
         sys_prompt = []
+        context = context[:]
```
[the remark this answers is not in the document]
dcasbol: Not "just". It's a soft copy, so that if I do `context.pop(1)` afterwards, it won't alter the original context.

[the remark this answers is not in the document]
JosephDavidsonKSWH: So it is!

utils/llm.py
```diff
@@ -67,81 +47,68 @@ def ensure_context_len(
...
+    while True:
+        num_tokens = response_len + count_tokens_for_model(model=model, context=sys_prompt + context)
+        if len(context) <= 1 or num_tokens <= max_len:
             break
-        messages.extend(message_pair)
-        context_tokens += message_tokens
-    messages.reverse()
-    context = sys_prompt + messages
-    # assert len(context) > 1, f"There are messages missing in the context:\n\n{context}"
-    return context, context_tokens
```
JosephDavidsonKSWH, line 98: All this was in here because Claude has some strong ideas about what the context should look like (i.e) no consecutive user, or assistant messages, and a user message has to come first. Maybe give it a test with Claude first?

dcasbol: Ah right! I sensed something like that, but yagni posessed me. I think I will just make sure that a user message comes first, and leave the more nuanced handling to the upper layers.

JosephDavidsonKSWH (approved):

utils/llm.py
dcasbol: Anthropic does not complain anymore about the first non-user message. I had to update haiku's alias though, since they updated it and deprecated the old one.

PR #121, opened 2024-07-12
LTM agent baseline with context limits

dcasbol:


ltm/agent.py
```diff
@@ -302,21 +305,17 @@ def reply(self, user_content: str, cost_callback: Callable[[float], Any] = None)
...
 
     def _completion(self, context: List[dict[str, str]], temperature: float, label: str,
                     cost_callback: Callable[[float], Any]) -> str:
-        # TODO: max_tokens is borked: https://github.com/BerriAI/litellm/issues/4439
-        response = completion(model=self.model, messages=context, timeout=self.config.timeout,
-                              temperature=temperature)
-        response_text = response['choices'][0]['message']['content']
+        response_text = ask_llm(
+            context, self.model, temperature=temperature, timeout=self.config.timeout, cost_callback=cost_callback,
+        )
```
JosephDavidsonKSWH, line 316: We can add `max_response_tokens` back in here since they fixed this in the new litellm. Might be worth it to remove the version requirement on litellm

ltm/agent.py
```diff
@@ -64,7 +64,12 @@ class LTMAgent:
 
     @property
     def save_name(self) -> str:
-        return f"{self.model}-{self.max_prompt_size}-{self.max_completion_tokens}"
+        model = self.model.replace("/", "-")
+        return f"{model}-{self.max_prompt_size}-{self.max_completion_tokens}"
+
+    @property
+    def max_input_tokens(self) -> int:
+        return int(0.9 * (self.max_prompt_size - self.max_completion_tokens))
```
JosephDavidsonKSWH, line 72: We should pick between `max_prompt_size` and `max_input_tokens` and use it consistently. That, or properly draw a distinction between them. Like `max_prompt_size` is how many tokens the model can deal with, and `max_input_tokens = max_prompt_size - max_response_tokens`

dcasbol: Yeah, they are not the same, and that is exactly the difference.

JosephDavidsonKSWH: Oh yeah, I completely glossed over the fact that I included the definition of `max_input_tokens` in my comment :P

JosephDavidsonKSWH (approved):

PR #123, opened 2024-07-24
Inserted Context LTM Baseline

JosephDavidsonKSWH:
Added new baseline agent: The `InsertedContextAgent`

Memories are saved as semantic chunks and as full interactions, when the agent replies:
- It generates keywords that describes the message
- It generates queries and query keywords for the semantic memory
- It gathers memories from the semantic memory, retrieved using, and filtered by, the query keywords.
- The memories are further filtered by an LLM,
- For each memory, the full interaction is retrieved and inserted into the context chronologically, avoiding repetitions.
- After the response is generated, the message and response are saved to the semantic memories using the keywords generated in the first step

Note: This version includes the embodied scratchpad, but it is not enabled.


ltm/scratchpad.py
```diff
@@ -49,6 +49,10 @@
 This is what the user information looks like now:
 {user_info}
 
+Try to answer these questions:
+- Does the message contain information that will be useful in future?
+- Is the information unimportant general knowledge, or useful user specific knowledge?
```
dcasbol, line 54: Maybe this part of the prompt is making it harder for the LLM to follow the environment's interface. In this step, it should really just answer 'yes' or 'no'. The idea sounds good though, so maybe adding a previous step in which you ask the agent this question, and right after it replies you can ask it the yes/no question.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+
+        fname = f"data/llm_calls/{self.session.session_id}/call_{self.llm_index}.json"
+        with open(fname, "w") as f:
+            colour_print("BlUE", f"Saving to: {fname}")
+            f.write(dump_context_s(context))
+            response_text = ask_llm(context, model=self.model, max_overall_tokens=self.max_prompt_size,
+                                    cost_callback=cost_callback, temperature=self.temperature)
+
+            f.write(f"\n\nResponse:\n{response_text}")
+            self.llm_index += 1
```
dcasbol, line 82: This is mostly a re-implementation of `_debug_actions`. You can probably just add `session_id` to the logic and reuse it.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+        for m in relevant_memories:
+            interaction = self.messages_from_memory(m)
+            if interaction not in full_interactions:
+                full_interactions.append(interaction)
+
+        for m in full_interactions:
+            if "trivia" in m[0].content:
+                colour_print("YELLOW", f"<*** trivia ***>")
+            else:
+                colour_print("YELLOW", f"{m[0].content}")
```
dcasbol, line 138: debug prints?

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+        shown_mems = 0
+        target_size = max_prompt_size - self.max_message_size
+
+        for interaction in full_interactions[::-1]:
+            user_message, assistant_message = interaction
+            future_size = current_size + token_counter(self.model, messages=[user_message.as_llm_dict(), assistant_message.as_llm_dict()])
+
+            # If this message is going to be too big, then skip it
+            if future_size > target_size or shown_mems == 100:
+                continue
```
dcasbol, line 163: It looks like this should be a `break` statement.

JosephDavidsonKSWH: I can split this out. Since everything being inserted are memories, I think that more memories are better, so I will prefer to skip memories that are too big, but insert any that can be inserted afterwards. Otherwise if a huge memory is retrieved, then the insertion will stop when we hit that memory and will end up with a gap that could otherwise be filled.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+            previous_interactions -= 1
+
+        # Add in memories up to the max prompt size
+        current_size = token_counter(self.model, messages=context)
+        shown_mems = 0
+        target_size = max_prompt_size - self.max_message_size
+
+        for interaction in full_interactions[::-1]:
+            user_message, assistant_message = interaction
+            future_size = current_size + token_counter(self.model, messages=[user_message.as_llm_dict(), assistant_message.as_llm_dict()])
```
dcasbol, line 159: Most models add tokens inbetween messages, so this way of counting tokens turns up slightly unreliable. If, however, you pass the whole context to the counter, the result will be most times exact.

ltm/inserted_context_agent.py
dcasbol: Ah, I see it now. I thought that once the condition was met it would be always true, but you're right. Nice solution.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+        Creates a new LTMAgentSession object and sets it as the current session.
+        :return: The new session object.
+        """
+        session_id = str(uuid.uuid4())
+        self.session = LTMAgentSession(session_id=session_id, m_history=[], i_dict={})
+        if not self.semantic_memory.is_empty():
+            self.semantic_memory.add_separator()
+        return self.session
+
+    def reply(self, user_message: str, agent_response: Optional[str] = None, cost_callback: Callable=None) -> str:
```
dcasbol, line 65: There are a couple of things you can leverage from the current baseline:
1. Set `self.now` at the very beginning of `reply`, to have a consistent reference timestamp.
2. Use `self._completion` to clean up the code a bit.
3. The function `td_format` to show relative timestamps.
(and probably other small things that I'm forgetting now)

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+
+SITUATION:
+{situation}
+
+Express your answer in this JSON: 
+[
+    {{
+        "number": int  // The number of the interaction.
+        "justification": string  // Why the interaction is or is not related to the situation.
+        "related": bool // Whether the interaction is related to the situation.
```
dcasbol, line 199: I think this is nice, and we should apply the idea more often:
1. Force the LLM into a decision.
2. Make it write a justification to have it think about it.
3. Allow it to retract.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+
+            queries_txt = "- " + "\n- ".join(queries)
+            passages = "\n----\n".join(memories_passages)
+            context = [make_user_message(prompt.format(passages=passages, situation=situation))]
+            # context = [make_system_message(prompt.format(passages=passages, queries=queries_txt, keywords=keywords))]
+
+            while True:
+                try:
+                    print("Attempting filter")
+                    with open(f"data/llm_calls/{self.session.session_id}/filter-{self.llm_index}-{call_count}.txt", "w") as f:
```
dcasbol, line 251: We should have this more standardized. I think that using the code that's currently in the baseline, plus session ID and some label should be enough. We could also use a timestamp instead of an index for the file name, to make the code simpler and also make it easier to find the right file.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+            self, message_history: list[Message], user_message: str, cost_cb: Callable[[float], None],
+            max_changes: int = 10, max_tries: int = 5, max_messages: int = 10, max_scratchpad_tokens: int = 1024,
+    ) -> dict:
+        """An embodied agent interacts with an extremely simple environment to safely update the scratchpad.
+        A rather small scratchpad (1k tokens) seems to work optimally."""
+        assert 0 < max_scratchpad_tokens < self.max_prompt_size // 2
+        ask_kwargs = dict(label="scratchpad", temperature=0, cost_callback=cost_cb, max_messages=max_messages)
+        scratchpad = deepcopy(self.scratchpad)
+        scratchpad_timestamps = self.scratchpad_ts
+        preview_tokens = (self.max_prompt_size // 2) - self.count_tokens(text=sp.to_text(self.scratchpad))
```
dcasbol, line 403: These two are not used, so remove them at your discretion, depending on whether you're planning to test them afterwards or not.

requirements.txt
```diff
@@ -6,7 +6,7 @@ faker~=20.1.0
 colorama~=0.4.6
 pytest
 time-machine==2.13.0
-tiktoken~=0.5.2
+tiktoken
```
dcasbol, line 9: About Antony's issue with the packages: we could simulate a fresh-new installation from time to time. I think the order of packages in the requirements file is important, and we can have it ordered in a way that conflicts don't show upd.

requirements.txt
```diff
@@ -16,7 +16,7 @@ Werkzeug==2.2.2  # It's crucial to specify this package's version to make flask
 flask==2.2.3
 browser-cookie3
 rouge_score==0.1.2
-litellm==1.41.13  # Unifies interface from most LLM APIs
+litellm # Unifies interface from most LLM APIs
```
dcasbol, line 19: I think it's better to have a version specified, so that we can control when the update is made. Otherwise, silent changes could happen. I could be fine with it, but at least I'd vote for adding a version before merging to main.

dcasbol:
I suggest the following before merging:
1. Pull current baseline.
2. Refactor and reuse/adapt as many elements from the baseline as possible, and as long they're helpful.
3. Discard unused elements.
4. Rename this agent to `agent.py` and remove or archive the previous one.

ltm/inserted_context_agent.py
```diff
@@ -0,0 +1,577 @@
...
+    max_prompt_size: int = 16384
+    is_local: bool = True
+    defined_kws: list = field(default_factory=list)
+    scratchpad: dict[str, str] = field(default_factory=dict)
+    scratchpad_ts: dict[str, float] = field(default_factory=dict)  # Tracks changes in user_info (timestamps)
+    llm_index: int = 0
+    costs_usd: float = 0.0
+    model: str = "together_ai/meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo"
+    temperature: float = 0.01
+    system_message = """You are a helpful AI assistant."""
```
dcasbol, line 45: `max_prompt_size` is defined twice.

ltm/inserted_context_agent.py
dcasbol: I see. I like the `reply_id/llm_call_id/label` structure. I also don't like the session id very much, but I reckon it's a good safety measure if you want to keep a previous run's logs.

JosephDavidsonKSWH: I have ditched the `session_id` and have replaced it with a timestamp

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+    def from_state_text(cls, state_text: str) -> 'LTMAgentSession':
+        """
+        Builds a session object given state text.
+        :param state_text: Text previously obtained using the state_as_text() method.
+        :return: A session instance.
+        """
+        state: dict = json.loads(state_text, cls=SimpleJSONDecoder)
+        session_id = state["session_id"]
+        m_history = state["history"]
+        i_dict = state["interactions"]
```
dcasbol, line 62: Here's a reminder of the bug I found:
```python
i_dict = {float(k): v for k, v in state["interactions"].items()}
```

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
-    prompt_callback: Callable[[str, str, list[dict], str], Any] = None
-    user_info: dict = field(default_factory=dict)
-    user_info_ts: dict[str, float] = field(default_factory=dict)  # Tracks changes in user_info (timestamps)
-    now: datetime.datetime = None  # Set in `reply` to keep a consistent "now" timestamp
+    semantic_memory: DefaultTextMemory = field(default_factory=AutoTextMemory.create)
+    max_prompt_size: int = 16384
+    is_local: bool = True
+    defined_kws: list = field(default_factory=list)
+    llm_call_idx: int = 0
+    costs_usd: float = 0.0
```
dcasbol, line 74: duplicated from agent wrapper (and not used)

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+        self.session.add_interaction((um, am))
+
+        return response_text
+
+    def keywords_for_message(self, user_message, cost_cb):
+
+        prompt = 'Create two keywords to describe the topic of this message:\n"{user_message}".\n\nFocus on the topic and tone of the message. Produce the keywords in JSON like: `["keyword_1", "keyword_2"]`\n\nChoose keywords that would aid in retriving this message from memory in the future.\n\nReuse these keywords if appropriate: {keywords}'
+
+        context = [make_system_message(prompt.format(user_message=user_message, keywords=self.defined_kws))]
+        while True:
```
dcasbol, line 129: I'd set a limit for the number of tries here.

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+        for k in keywords:
+            if k not in self.defined_kws:
+                self.defined_kws.append(k)
+
+        print(f"Interaction keywords: {keywords}")
+        return keywords
+
+    def create_context(self, user_message, max_prompt_size, previous_interactions, cost_cb):
+
+        stamped_user_message = str(self.now) + ": " + user_message
```
dcasbol, line 150: This timestamp representation is different from the one in line 193, and also from the one used in the previous baseline:
```text
[2024-07-31 12:00:56 (7 minutes and 4 seconds ago)]
<user_message_here>
```

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+                colour_print("YELLOW", f"{m[0].content}")
+
+        # Add the previous messages
+        final_idx = self.session.message_count - 1
+        while previous_interactions > 0 and final_idx > 0:
+
+            # Agent reply
+            context.insert(1, self.session.by_index(final_idx).as_llm_dict())
+            # User message
+            context.insert(1, self.session.by_index(final_idx-1).as_llm_dict())
```
dcasbol, line 168: These messages are not timestamped. In the previous baseline they were timestamped. I guess you didn't notice it because you're not adding any to the context, but it seems right to continue the pattern.

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+            context.insert(1, assistant_interaction.as_llm_dict())
+
+            ts = datetime.datetime.fromtimestamp(user_interaction.timestamp)
+            et_descriptor = f"{str(ts)[:-7]} ({td_format(self.now - ts)}) "
+            context.insert(1, user_interaction.as_llm_dict())
+            context[1]["content"] = et_descriptor + context[1]["content"]
+
+            shown_mems += 1
+
+            current_size = future_size
```
dcasbol, line 199: Just a couple of minor optimizations:
1. I find `reversed(relevant_interactions)` more readable, and it also does not make a copy of the list (`[::-1]` does).
2. `if show_mems >= 100` could go right after increasing the counter in line 197.

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+
+            for interaction, keywords in zip(interactions_to_filter[start_idx:end_idx], interaction_keywords[start_idx:end_idx]):
+                um, am = interaction
+                memories_passages.append(f"[MEMORY NUMBER {memory_counter} START].\n (User): {um.content}\n(You): {am.content}\nKeywords: {keywords}\n[MEMORY NUMBER {memory_counter} END]")
+                memory_counter += 1
+
+            passages = "\n\n------------------------\n\n".join(memories_passages)
+            context = [make_user_message(prompt.format(passages=passages, situation=situation))]
+
+            while True:
```
dcasbol, line 267: set a limit

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+{{
+    "queries": array, // An array of strings: 2 descriptive search phrases, one general and one specific
+    "keywords": array // An array of strings: 1 to 3 keywords that can be used to narrow the category of memories that are interesting. 
+}}"""
+
+        # Now create the context for generating the queries
+        context = [make_user_message(prompt.format(user_message=user_message, time=self.now, keywords=self.defined_kws))]
+        all_retrieved_memories = []
+        query_keywords = []
+        while True:
```
dcasbol, line 317: limit number of tries

ltm/agent.py
```diff
@@ -1,102 +1,408 @@
...
+        #     if "trivia" in kw:
+        #         trivia_skip = True
+        #
+        # if trivia_skip:
+        #     llm_filtered_interactions, _ = self.interactions_from_retrieved_memories(keyword_filtered_mems)
+        # else:
+        #     llm_filtered_interactions = self.llm_memory_filter(keyword_filtered_mems, query_dict["queries"], cost_cb)
+
+        # TODO: ....And comment this one out
+        llm_filtered_interactions = self.llm_memory_filter(keyword_filtered_mems, query_dict["queries"], cost_cb)
```
dcasbol, line 368: Could it be good to make this depend on the debug level, instead of commenting it in/out?

dcasbol (approved):

ltm/agent.py
JosephDavidsonKSWH: The actually are, just after they are added to the context see the lines below. We update the one at context[1], which is the user message.

dcasbol: I see that all user messages before these interactions (added in the loop after this) are timestamped, but if `previous_interactions > 0`, the most recent interactions added in this loop don't have a timestamp. Am I missing something?

JosephDavidsonKSWH: Oh yeah, you are correct here, I will actually tidy it up and unify the two

dcasbol (approved):

PR #124, opened 2024-07-26
Improve progress bar & reword test questions

dcasbol:


runner/progress.py
```diff
@@ -52,21 +60,39 @@ def update_stats(self):
...
-            progress /= max(sum(total), 1)
+            progress /= max(total_tokens, 1)
         self._progressbar["value"] = int(100 * progress)
+
+        # First row shows the total score so far and the tests completed
+        num_finished = 0
+        total_score = 0
+        for scores in self._scores.values():
+            for s in scores:
+                total_score += s / self._num_reps
```
JosephDavidsonKSWH, line 80: If I am reading this right, `num_reps` shouldn't be universal. Sure it is a part of the config, but that number can be overridden at the test level (again in the config). So while there may be nominally 3 repetitions of a test, I could have it so that there is only one example of ChapterBreak for example

dcasbol: That's absolutely right. I think the same, but I didn't know how to get around it... oh, maybe I have. Give me a minute 😬 🤞🏾

JosephDavidsonKSWH (approved):

runner/progress.py
```diff
@@ -52,21 +62,39 @@ def update_stats(self):
...
-            progress /= max(sum(total), 1)
+            progress /= max(total_tokens, 1)
         self._progressbar["value"] = int(100 * progress)
+
+        # First row shows the total score so far and the tests completed
+        num_finished = 0
+        total_score = 0
+        for dataset_name, scores in self._scores.items():
+            for s in scores:
+                total_score += s / self._repetitions[dataset_name]
```
dcasbol, line 82: this should fix it @JosephDavidsonKSWH

JosephDavidsonKSWH (approved):

PR #126, opened 2024-08-01
refactor(charlie_interface): Replace browser-based cookie extraction …

airobinnet:
…with direct login via requests.session and modularize actions

- Removed `browser_cookie3` dependency
- Added methods for login, creating chat tabs, updating memory settings, and system prompts
- Updated default values for credentials and endpoint configurations
- Enhanced error handling and added comments for clarity

### DIFF
--- a/model_interfaces/charlie_interface.py
+++ b/model_interfaces/charlie_interface.py
@@ -8,8 +8,8 @@ from model_interfaces.interface import ChatSession
 
 @dataclass
 class CharlieMnemonic(ChatSession):
-    max_prompt_size: int = 4160  # match with input_tokens in update_memory_settings
-    chat_id: str = "Benchmark Chat"
+    max_prompt_size: int = 9913  # match with input_tokens in update_memory_settings
+    chat_id: str = "benchmarkchat"
     endpoint: str = "http://localhost:8002"  # default local endpoint
     username: str = "admin"  # default username
     password: str = "admin"  # default password
@@ -19,6 +19,7 @@ class CharlieMnemonic(ChatSession):
     system_prompt: str = (
         "You are answering questions for an automated benchmark, don't ask the user if you should save information, just save it."
     )
+    reset_on_start: bool = True
 
     @property
     def name(self):
@@ -28,11 +29,8 @@ class CharlieMnemonic(ChatSession):
         super().__post_init__()
         self.session = requests.Session()
         self.login()
-        self.create_chat_tab()
-        self.update_memory_settings()
-        self.update_max_tokens()
-        self.load_initial_settings()
-        self.update_system_prompt(self.system_prompt)
+        if self.reset_on_start:
+            self.reset()
 
     def login(self):
         login_url = f"{self.endpoint}/login/"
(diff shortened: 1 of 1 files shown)


model_interfaces/charlie_interface.py
```diff
@@ -1,143 +1,149 @@
...
+        if response.status_code != 200:
+            raise ValueError("Failed to update system prompt setting.")
+        self.system_prompt = new_system_prompt
+
+    def create_chat_tab(self):
+        create_chat_tab_url = f"{self.endpoint}/create_chat_tab/"
+        chat_tab_data = {
+            "username": self.username,
+            "chat_id": self.chat_id,
+            "chat_name": f"{self.chat_id}",
```
dcasbol, line 61: I wonder if there's any motivation behind the use of the f-string, and what would happen if a chat with the same ID already exists.

model_interfaces/charlie_interface.py
```diff
@@ -1,143 +1,149 @@
...
-
-    def reply(self, user_message: str, agent_response: Optional[str] = None) -> str:
-        headers = {
-            "Content-Type": "application/json",
-            "Cookie": f"session_token={self.token}",
+        response = self.session.post(update_settings_url, json=update_data)
+        if response.status_code != 200:
+            raise ValueError("Failed to update max_tokens setting.")
+
+    # Charlie mnemonic has a lot of memory settings that need to be updated, performance varies greatly based on these settings
```
dcasbol, line 78: In the future, it would be nice to have a default memory distribution. As a user, I'd just set the max. input size, or the model to use, and the different sections would adjust acordingly.

model_interfaces/charlie_interface.py
```diff
@@ -1,143 +1,149 @@
...
-        )
-
+        delete_data_url = f"{self.endpoint}/delete_data_keep_settings/"
+        response = self.session.post(delete_data_url)
+        if response.status_code != 200:
+            raise ValueError("Failed to reset user data.")
 
     def load(self):
         # Charlie mnemonic is web based and so doesn't need to be manually told to resume a conversation
+        # Use the chat_id to continue a conversation
```
dcasbol, line 146: I'd avoid resuming with charlie until we have a checkpoint-based persistence mechanism. The benchmark relies on setting the agent back to a specific point, while the web's state might be anywhere between two checkpoints.

dcasbol (approved):

model_interfaces/charlie_interface.py
airobinnet: f-string is not per > I wonder if there's any motivation behind the use of the f-string
F-string is not really needed here, initially I did something in the line of `"chat_name": f"Benchmark {self.chat_id}",`
but then removed the 'Bechmark ' part.

> and what would happen if a chat with the same ID already exists.
Would throw an error, I'll add something to avoid this error.se needed here,

model_interfaces/charlie_interface.py
airobinnet: Agree, these settings are probably confusing for the user.

PR #130, opened 2024-08-08
task memory for remembering things to do outside of the context

JosephDavidsonKSWH:
Added task memories
- Task memories are parsed from user messages + above context that is relevant to the message.
- If a task is deemed to have been specified by the user, a JSON object is created and added to a list.
- Each step, all tasks are checked to see if their trigger condition is met (either # of messages, time, or message content)
- If so, the agent is prompted to perform the task alongside acknowledging the users' message.
- Tasks that have been triggered are cleaned up afterwards. 

Other changes:
- Adjusted the situation prompt to give a little more context.
- Reply behaviour changes based on whether there is an expected response from the agent.

### DIFF
--- a/ltm/agent.py
+++ b/ltm/agent.py
@@ -151,7 +151,7 @@ Contains instructions for a future task which is not already in the list of futu
 - Does the user tell the agent that it should perform some kind of task in the future, along with some trigger condition for that task?
 - Is the task already in the list?
 
-Respond with your reasoning as to whether there has been a new future task discussed and then with either "Yes." or "No." written exactly."""
+Respond with your reasoning as to whether there has been a new future task discussed and with either "Should Update: Yes." or "Should Update: No." written exactly at the very end of your response."""
 
         new_task_def = """Create a new task in JSON format.
 A task contains two things, a trigger condition and a payload.
@@ -180,7 +180,9 @@ Follow this format:
         log_llm_call(self.run_name, self.save_name, self.debug_level, label=f"should_update-{self.llm_call_idx}")
         colour_print("YELLOW", f"Should update: {response_text}")
 
-        if "No." in response_text:
+        if "Should Update: No." in response_text:
+            return
+        elif "Should Update: Yes." not in response_text:
             return
 
         # Produce a new one
@@ -228,7 +230,7 @@ Follow this format:
         if cond == "MESSAGE_COUNT":
             # Decrement the message counter and ready it to be triggered
             task["trigger_details"]["trigger_value"] -= 1
-            task["trigger_details"]["triggered"] = val <= 0
+            task["trigger_details"]["triggered"] = task["trigger_details"]["trigger_value"] <= 0
 
         elif cond == "MESSAGE_CONTENT":
             user_message = context[-1]["content"].lower()
--- a/ltm/agent.py
+++ b/ltm/agent.py
@@ -1,6 +1,8 @@
 import json
 import re
+import time
 import uuid
+from copy import deepcopy
 from dataclasses import dataclass, field
 import datetime
 from math import ceil
@@ -15,7 +17,7 @@ from ltm.memory.hybrid_memory import HybridMemory
 from ltm.utils.config import Config
 from model_interfaces.base_ltm_agent import Message
 
-from utils.llm import make_system_message, make_user_message, ask_llm, log_llm_call
+from utils.llm import make_system_message, make_user_message, ask_llm, log_llm_call, make_assistant_message
 from utils.text import td_format
 from utils.ui import colour_print
 
@@ -38,6 +40,7 @@ class InsertedContextAgent:  # ? worth adding most of this to the ltm.utils.conf
     now: datetime.datetime = None  # Set in `reply` to keep a consistent "now" timestamp
     run_name: str = ""
     num_tries: int = 5
+    task_memory: list = field(default_factory=list)
 
     @property
     def save_name(self) -> str:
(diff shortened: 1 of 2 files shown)


ltm/agent.py
```diff
@@ -66,15 +68,39 @@ def new_session(self) -> str:
         return session_id
 
     def reply(self, user_message: str, agent_response: Optional[str] = None, cost_callback: Callable = None) -> str:
+        perform_task = """
+
+Here are some additional tasks which you are to perform right now:
+{task}
+
+These tasks contains two things: a trigger condition, and a payload.
+The trigger conditions have been met for all task. So they can be safely disregarded. Now all you need to do is perform the tasks described in the "payload" fields.
```
dcasbol, line 77: Could you maybe just feed it the task descriptions, so that you can skip this explanation and have the agent directly perform the tasks?

ltm/agent.py
```diff
@@ -66,15 +68,39 @@ def new_session(self) -> str:
...
+
+        # Update task memory
+        self.add_task(deepcopy(context), cost_callback, user_message)
+
+        # Get any tasks that might be due now.
+        extra_tasks = self.progress_and_get_extra_tasks(context)
+        if len(extra_tasks) > 0:
+            context[-1]["content"] += perform_task.format(task=json.dumps(extra_tasks, indent=2))
+
+        response_text = ask_llm(context, model=self.model, max_overall_tokens=self.max_prompt_size, cost_callback=cost_callback, temperature=self.temperature)
```
dcasbol, line 103: This LLM call is not logged.

ltm/agent.py
JosephDavidsonKSWH: I tired that first, but it gets confused and says something like `Okay I will append the quote from x after the next n responses` Giving it less information helps focus it a little more.

ltm/agent.py
```diff
@@ -96,6 +125,121 @@ def sanitize_string(s: str) -> str:
...
+
+        context.append(make_user_message(task_should_add.format(future_tasks=json.dumps(self.task_memory, indent=2), user_message=user_message)))
+        response_text = ask_llm(context, model=self.model, max_overall_tokens=self.max_prompt_size, cost_callback=cost_callback, temperature=self.temperature)
+        log_llm_call(self.run_name, self.save_name, self.debug_level, label=f"should_update-{self.llm_call_idx}")
+        colour_print("YELLOW", f"Should update: {response_text}")
+
+        if "Should Update: No." in response_text:
+            return
+        elif "Should Update: Yes." not in response_text:
+            return
```
dcasbol, line 185: This seems a bit too rigid. I usually ask the LLM to give me a single-word 'yes' or 'no' answer, and then I just do `if "no" in response_text.lower()` to ignore possible casing or formatting variations.

ltm/agent.py
```diff
@@ -96,6 +125,121 @@ def sanitize_string(s: str) -> str:
...
+        if new_task["trigger_details"]["trigger_condition"] == "TIME":
+            new_task["trigger_details"]["trigger_value"] = time.mktime(datetime.datetime.strptime(new_task["trigger_details"]["trigger_value"], "%Y-%m-%d_%H:%M").timetuple())
+
+        self.task_memory.append(new_task)
+
+    def remove_completed_tasks(self):
+
+        for task in self.task_memory:
+            if task["trigger_details"]["triggered"]:
+                self.task_memory.remove(task)
```
dcasbol, line 220: `remove` looks for a perfect match, and `task` is a dictionary with many details. A safer way would be:
```python
self.task_memory = [
    t for t in self.task_memory
    if not t["trigger_details"]["triggered"]
]
```

ltm/agent.py
```diff
@@ -96,6 +125,121 @@ def sanitize_string(s: str) -> str:
...
+            user_message = context[-1]["content"].lower()
+            task["trigger_details"]["triggered"] = val in user_message
+
+        elif cond == "TIME":
+            task["trigger_details"]["triggered"] = self.now.timestamp() > val
+
+        if task["trigger_details"]["triggered"]:
+            print(f"{cond}: Task triggered: {json.dumps(task, indent=2)}")
+
+        return task["trigger_details"]["triggered"]
```
dcasbol, line 241: This looks better and it's easier to read than `process_new_task`. You could maybe do `details = task["trigger_details"]` too.

ltm/agent.py
JosephDavidsonKSWH: Here the rigidity is the point. The agent has a bunch of free text to think though the points and a  look for "no" matches a ton of words. The alternative is to have it answer in JSON.

ltm/agent.py
dcasbol: I see. That last sentence put me off though. I'd say that you're giving it more information than if just giving the task descriptions.

ltm/agent.py
dcasbol: Ah right. So the point is to have the LLM write some reasoning and let that decision be the last part of it 👌🏾

dcasbol:
One thing that I do and that I'd encourage you to do as well is to rely on a single method for performing the LLM calls and logs. This is how my method looks:
```python
    def _completion(self, context: LLMContext, temperature: float, label: str, cost_callback: CostCallback) -> str:
        response_text = ask_llm(
            context, self.model,
            temperature=temperature, cost_callback=cost_callback, max_overall_tokens=self.max_prompt_size,
        )
        log_llm_call(self.run_name, self.agent_log_id, self.debug_level, label=label)
        return response_text
```
Where `self.agent_log_id` is `f"{init_timestamp} - {self.save_name}"`.

dcasbol (approved):

ltm/agent.py
JosephDavidsonKSWH: Its really not a lot of information, here is a live example:
```
2024-08-09 14:55:53 (just now) I am hppy today!

Here are some additional tasks which you are to perform right now:
[
  {
    "trigger_condition": "The 6th response after the current message.",
    "payload": "Append the quote 'Be yourself; everyone else is already taken.' - Oscar Wilde to the response."
  }
]

These tasks contains two things: a trigger condition, and a payload.
The trigger conditions have been met for all task. So they can be safely disregarded. Now all you need to do is perform the tasks described in the "payload" fields. 

You should append the result of your task to your reply to the query above

Append the results of your tasks to your current response.
```
 I see the explanation as insurance, because sometimes the payload is phrased incorrectly e.g (quote xxx after the nth response) Which usually confuses the agent. This setup is the one that has worked the best so far.

PR #132, opened 2024-08-12
new benchmarks

JosephDavidsonKSWH:


configurations/published_benchmarks/benchmark-v3-120k-#2.yml
```diff
@@ -0,0 +1,49 @@
+config:
+  debug: True
+  run_name: "Benchmark 3 - 120k #3"
```
dcasbol, line 3: `#3` is a typo, I guess.

data/tests/Benchmark 3 - 120k #3/definitions/ChapterBreak/47583370_7.def.json
```diff
@@ -0,0 +1,53 @@
+{
```
dcasbol, line 1: It seems that ChapterBreak has the very same samples than the `#1` version. There should be at least 4 samples reviewed, and we might need to review another two. These are of especial relevance, since they show up at the very beginning.

data/tests/Benchmark 3 - 120k #3/definitions/config.yml
```diff
@@ -0,0 +1,37 @@
+config:
+  debug: true
+  incompatibilities:
+  - - names
+    - name_list
+  - - locations
+    - locations_directions
+  run_name: 'Benchmark 3 - 120k #3'
```
dcasbol, line 8: The typo was copied over to the definitions.

PR #133, opened 2024-08-12
Agent performs time-forwarding. Falls back to regular waits.

dcasbol:


model_interfaces/interface.py
```diff
@@ -64,3 +69,26 @@ def load(self):
 
     def token_len(self, text: str) -> int:
         return count_tokens_for_model(text=text)
+
+    def forward_time(self, seconds: float):
```
JosephDavidsonKSWH, line 73: I am unclear on how moving the time traveling functionality to the model interface accomplishes what we want. Won't an override of this in the model implementation break this functionality?

dcasbol: An override should only happen when needed. Some examples:
1. Charlie Mnemonic will perform an API call to communicate the time jump to the server.
2. An agent that doesn't implement any measure for this can just set `self.time_travel = False`.
3. An agent that has a different way of handling timestamps can implement its own method to simulate these time travels.

dcasbol: Ah... I see now what you mean. We should perform the time jumps **in addition** to letting the agents implement their own methods. Otherwise, benchmark and agent will get out of sync. I will fix that.

JosephDavidsonKSWH (approved):

PR #145, opened 2024-08-28
Retrieval Eval. Evaluations for all benchmarks

FKGSOFTWARE:


.gitignore
```diff
@@ -40,4 +40,8 @@ not_for_github/auto_scorer/sheet_to_local/*
 #RDB stuff
 win-RDB-venv/
 ltm_agent.db
-.vscode
\ No newline at end of file
+.vscode
+
+#Eval Stuff
+retrieval_evaluator/comparison_data/*
+retrieval_evaluator/evaluation_outputs/*
```
dcasbol, line 47: This could probably go under `data`, which is ignored by default.

retrieval_evaluator/retrieval_evaluator.py
```diff
@@ -0,0 +1,558 @@
...
+            return []
+        except json.JSONDecodeError:
+            self.logger.error(f"Error decoding JSON from file: {file_path}")
+            return []
+
+    def is_trivia_memory(self, memory):
+        return "Here are some trivia questions and answers for you to process." in memory.get('query', '')
+
+    def fuzzy_match(self, query1, query2, threshold=85):
+        return fuzz.ratio(query1.lower(), query2.lower()) >= threshold
```
dcasbol, line 74: You mentioned that this was giving you some issues. In other parts of the benchmark I've used the `standardise` and `rouge_l` functions from `utils.text` to accomplish a similar goal, but I believe that ROUGE is even more strict than L-distance.

.gitignore
FKGSOFTWARE: Moved data related to data and the script under Utils

retrieval_evaluator/retrieval_evaluator.py
FKGSOFTWARE: I will have a look into this

utils/retrieval_evaluator.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def plot_original_results(self, results, summary, output_path):
+        df = pd.DataFrame(results)
+        df['Entry'] = range(1, len(df) + 1)
+
+        plt.style.use('default')
+        fig = plt.figure(figsize=(24, 16))
+        gs = GridSpec(3, 2, height_ratios=[1, 0.5, 3], width_ratios=[1.2, 0.8])
+
+        colors = {'True_relevant': '#4CAF50', 'Irrelevant': '#FFC107', 'Trivia': '#9C27B0', 'Missing_Relevant': '#F44336'}
```
dcasbol, line 161: You've used `True_**r**elevant` and `Missing_**R**elevant`, and in other places I've seen you accessing the dict with `colors[category.capitalize()]`. This would work in the first case but not in the second.

utils/retrieval_evaluator.py
```diff
@@ -0,0 +1,559 @@
...
+
+            retrieved = [r['retrieved_memories_count'] for r in results]
+            filtered = [r['filtered_memories_count'] for r in results]
+            t_stat, p_value = stats.ttest_rel(retrieved, filtered)
+            f.write(f"\nPaired t-test results (Retrieved vs Filtered):\n")
+            f.write(f"t-statistic: {t_stat:.4f}\n")
+            f.write(f"p-value: {p_value:.4f}\n")
+
+    def capture_comparison_data(self, query: str, retrieved_memories: List[Union[Dict[str, Any], Any]], filtered_memories: List[Union[Dict[str, Any], Any]]):
+        comparison_data = {
```
JosephDavidsonKSWH, line 475: Don't know if you can make it work, but the current benchmark uses ~3 queries to get the retrieved memories. Are you able to adapt the query string to a list?

Right now I am just `"\n".join()`-ing the queries.

data/retrieval_evaluator/eval_readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+evaluator = RetrievalEvaluator()
+```
+
+## 2. Capturing Comparison Data
+During your benchmark run, capture data for each query:
+```
+query = "Your query here"
+relevant_memories = [...]  # List of retrieved memories
+filtered_memories = [...]  # List of filtered memories
```
dcasbol, line 16: It would be nice to include the specification for the memories.

data/retrieval_evaluator/eval_readme.md
```diff
@@ -0,0 +1,37 @@
...
+Note: You don't need to specify the benchmark version at this stage.
+
+## 3. Generating Output
+After your benchmark run is complete, generate the output for a specific benchmark version:
+   `evaluator.output("Dev Benchmark 1 32k - 2 Examples")`
+
+## 4. Output Location
+The evaluation results will be saved in:
+
+Plots and result files: retrieval_evaluator/evaluation_outputs/evaluation_{benchmark_version}_{timestamp}/
```
dcasbol, line 29: Does this still work this way? I would expect it to require a file path to save to and, if not provided, just show it on screen.

data/retrieval_evaluator/eval_readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+Log file: retrieval_evaluator/logs/evaluation_{benchmark_version}.log
+
+## Important Notes
+
+Ensure that the reference data files exist in retrieval_evaluator/dev_bench_reference_data/ directory.
```
dcasbol, line 35: From what you told me, I guess these files aren't required anymore, right?

ltm/agent.py
```diff
@@ -459,7 +463,7 @@ def get_relevant_memories(self, user_message, cost_cb):
                 if r_mem.relevance > 0.6:
                     relevance_filtered_mems.append(r_mem)
 
-        llm_filtered_interactions = self.llm_memory_filter(relevance_filtered_mems, query_dict["queries"], cost_cb)
+        llm_filtered_interactions = self.llm_memory_filter(relevance_filtered_mems, query_dict["queries"], user_message, cost_cb)
```
dcasbol, line 466: Wouldn't be better to call the evaluator after this line, rather than adding the user message as a parameter just for calling it inside? Here you already have `llm_filtered_interactions` and `relevance_filtered_mems`, so in principle you could call it right after performing the filtering.

runner/scheduler.py
```diff
@@ -481,6 +481,21 @@ def run(self):
...
+                example.reference_data_from_log(list(self.master_log.test_events(example.unique_id))))
+
+        reference_data_path = RETRIEVAL_REFERENCE_DIR.joinpath(self.config.run_name, "reference_data.json")
+        reference_data_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(reference_data_path, 'w') as f:
+            json.dump(reference_data, f, indent=2)
+
+        if self.agent.retrieval_evaluator:
+            self.agent.retrieval_evaluator.output(self.config.run_name)
+
```
dcasbol, line 498: If this doesn't take long and can be performed with the data available in the master log, it could be better to just have it extracted somewhere else, probably in a dedicated script? That way we can also spare saving that file and perform the eval on any benchmark log.

utils/retrieval_evaluator.py
```diff
@@ -0,0 +1,627 @@
...
+    def read_data(self):
+        if not self.comparison_data_file.exists():
+            print(f"Comparison data file not found at {self.comparison_data_file}")
+            return []
+        with open(self.comparison_data_file, 'r') as f:
+            return json.load(f)
+
+class RetrievalEvaluator:
+    def __init__(self):
+        self.project_root = Path(__file__).parent.parent / "data" / "retrieval_evaluator"
```
dcasbol, line 71: should use `DATA_DIR`

utils/retrieval_evaluator.py
```diff
@@ -0,0 +1,627 @@
...
+            with open(file_path, 'r') as file:
+                return json.load(file)
+        except FileNotFoundError:
+            self.logger.warning(f"File not found: {file_path}")
+            return []
+        except json.JSONDecodeError:
+            self.logger.error(f"Error decoding JSON from file: {file_path}")
+            return []
+
+    def is_trivia_memory(self, memory: Dict[str, Any]) -> bool:
```
dcasbol, line 117: Could we use the `filler` flag for this?

utils/util_math.py
```diff
@@ -0,0 +1,31 @@
+from math import sqrt
```
dcasbol, line 1: The sole purpose of this file was to avoid adding an unnecessary dependency on numpy, but the memory analysis already requires it, so you can safely delete this and use the numpy methods where needed. As a side note, `sqrt = lambda x: x ** 0.5`.

utils/retrieval_evaluator.py
```diff
@@ -0,0 +1,627 @@
...
+        with open(self.comparison_data_file, 'r') as f:
+            return json.load(f)
+
+class RetrievalEvaluator:
+    def __init__(self):
+        self.project_root = Path(__file__).parent.parent / "data" / "retrieval_evaluator"
+        self.dev_bench_reference_data_path = RETRIEVAL_REFERENCE_DIR
+        self.comparison_data_path = self.project_root / "comparison_data"
+        self.logs_path = self.project_root / "logs"
+        self.evaluation_outputs_path = self.project_root / "evaluation_outputs"
```
dcasbol, line 75: I think it would be very handy to be able to do something like `python evaluate_retrieval.py "Run name" "Agent name"` and have the code grab the full log and all, instead of having to go through a file.

ltm/agent.py
JosephDavidsonKSWH: So `relevance_filtered_mems` are memory chunks rather than interactions. What is required is that we call `interactions_from_retrieved_memories` inside the function to get the interactions that are then filtered. And we actually store those and the filtered interactions in the evaluator. 

I can refactor this to be a little more single responsibility, but it will add more parameters to the function, not fewer.

utils/retrieval_evaluator.py
JosephDavidsonKSWH: This would only work if all the memory retrievals were saved by default. While we are able to generate the reference data from the test defs and he master log for a run, we only get the retrieval stats if the agent gives them to us.

Since we treat the agents as a black box, there is no guarantee that this can work, and I don't want to institute boilerplate for them.

dcasbol: I see. I oversimplified it in my head.

ltm/agent.py
dcasbol: Understood. In that case, It's fine like it is I think.

ltm/agent.py
```diff
@@ -308,7 +310,7 @@ def create_context(self, user_message, max_prompt_size, previous_interactions, c
 
         return context
 
-    def llm_memory_filter(self, memories, queries, cost_cb):
+    def llm_memory_filter(self, interactions_to_filter, interaction_keywords, queries, cost_cb):
```
dcasbol, line 313: input and output typings would be nice

runner/master_log.py
```diff
@@ -2,6 +2,7 @@
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+from os.path import exists
```
dcasbol, line 5: not used anywhere

dcasbol (approved):

PR #147, opened 2024-09-04
Extended thinking pipeline: comments each message, then adds a global comment.

dcasbol:


ltm/agent.py
```diff
@@ -259,56 +283,39 @@ def keywords_for_message(self, user_message, cost_cb):
...
         return keywords
 
-    def create_context(self, user_message, max_prompt_size, previous_interactions, cost_cb):
-        context = [make_system_message(self.system_message), make_user_message(
-            f"{str(self.now)[:-7]} ({td_format(datetime.timedelta(seconds=1))}) " + user_message)]
-        relevant_interactions = self.get_relevant_memories(user_message, cost_cb)
-
-        # Get interactions from the memories
-        for m in relevant_interactions:
-            if "trivia" in m[0].content:
```
JosephDavidsonKSWH, line 269: This statement here is incredibly useful for at-a-glance sanity checking in-progress runs. If the memories are not what you expect then you can investigate further or terminate the run. Can we wrap some of this as a debug statement?

ltm/agent.py
```diff
@@ -6,32 +6,38 @@
...
+    LLMContext, count_tokens_for_model
+from utils.text import stamp_content
 from utils.ui import colour_print
 
 
+CostCallback = Callable[[float], None]
+
+
+def init_timestamp_factory():
+    return datetime.datetime.now().strftime("%F %T")
```
JosephDavidsonKSWH, line 29: What is the benefit of this local function over the definition in `__post_init__:69`?

ltm/agent.py
```diff
@@ -76,43 +93,50 @@ def reply(self, user_message: str, agent_response: Optional[str] = None, cost_ca
...
+        context = [make_system_message(self.system_message)] + memories + [
+            make_user_message(stamp_content(user_message, self.now, dt=self.now))
+        ]
+        self.add_task(deepcopy(context), user_message)
 
         # Get any tasks that might be due now.
         extra_tasks = self.progress_and_get_extra_tasks(context)
-        if len(extra_tasks) > 0:
-            context[-1]["content"] += perform_task.format(task=json.dumps(extra_tasks, indent=2))
+        if len(extra_tasks) > 0 or agent_response is None:
```
JosephDavidsonKSWH, line 117: This breaks prospective memory tests as it will only perform a task on messages where there are no trivia. The agent has to be able to perform tasks on any message.

ltm/agent.py
```diff
@@ -259,56 +283,39 @@ def keywords_for_message(self, user_message, cost_cb):
...
-        relevant_interactions = self.get_relevant_memories(user_message, cost_cb)
-
-        # Get interactions from the memories
-        for m in relevant_interactions:
-            if "trivia" in m[0].content:
-                colour_print("YELLOW", f"<*** trivia ***>")
-            else:
-                colour_print("YELLOW", f"{m[0].content}")
+    def collect_memories(self, user_message: str, previous_interactions: int) -> LLMContext:
+        relevant_interactions = self.get_relevant_memories(user_message)
```
JosephDavidsonKSWH, line 287: When counting tokens, this function does not take the size of the current user message into account.

ltm/agent.py
dcasbol: Just that `fields` can take it, but actually that makes it a worse option. I'll revert it.

ltm/agent.py
dcasbol: It will perform the task in both cases, but maybe this is a sign that the way it's put is maybe unintuitive. I can re-write it in a more sensible way.

JosephDavidsonKSWH: Oh yeah, its an `or` there!

ltm/agent.py
dcasbol: Ah, yes it does. I created a property `max_input_size`, which is `max_prompt_size - max_message_size`.

runner/scheduler.py
```diff
@@ -362,8 +362,9 @@ def run_tests(self):
         if len(self.master_log.log) == 0 and not self.config.isolated:
             self.send_message("", SendMessageAction(message=(
                 "I am going to subject you to a Long-Term Memory benchmark. In the following, I will be giving you "
-                "different kinds of information and I expect you to answer extremely briefly, only providing the "
-                "responses that you are required to provide. Otherwise, provide just short confirmations. Understood?"
+                "different kinds of information and I expect you to answer extremely briefly and not making any "
```
JosephDavidsonKSWH, line 365: `not asking any `...

utils/llm.py
```diff
@@ -196,13 +193,13 @@ def log_llm_call(run_name: str, agent_name: str, debug_level: int, label: str =
...
     save_dir.mkdir(parents=True, exist_ok=True)
 
-    # Sanitize timestamp
-    sanitized_timestamp = datetime.now().strftime("%Y-%m-%d_%H_%M_%S_%f")
-
     # Write content of LLM call to file
     context = _llm_debug_params.pop("messages")
     response_text = _llm_debug_params.pop("response")
-    save_path = save_dir.joinpath(f"{sanitized_timestamp}{'-' + label if label is not None else ''}.txt")
+    save_name = datetime.now().strftime("%Y-%m-%d_%H_%M_%S_%f")
```
JosephDavidsonKSWH, line 199: Minor thing, but maybe the format `"%Y-%m-%d_%H-%M-%S-%f"` would be more legible? The switch to underscores starting with the separator between the date and time makes it a little hard to parse at times.

runner/scheduler.py
dcasbol: ah! good catch

ltm/agent_think.py
```diff
@@ -0,0 +1,43 @@
...
+from utils.ui import colour_print
+from utils.llm import LLMContext, make_user_message, make_assistant_message
+
+
+@dataclass
+class LTMReflexionAgent(LTMAgent):
+
+    def _truncated_completion(self, context: LLMContext, max_messages: int = None, **kwargs) -> str:
+        max_messages = max_messages or len(context) + 1
+        while len(context) + 1 > max_messages or self.count_tokens(messages=context) > self.max_input_tokens:
```
JosephDavidsonKSWH, line 13: Are any of the methods in this subclass ever used?

runner/run_benchmark.py
```diff
@@ -12,23 +11,13 @@
...
-from model_interfaces.length_bias_agent import LengthBiasAgent
 from model_interfaces.interface import ChatSession
 from model_interfaces.llm_interface import LLMChatSession, TimestampLLMChatSession
-from model_interfaces.ltm_agent_wrapper import LTMAgentWrapper
-from model_interfaces.cost_estimation import CostEstimationChatSession
-from model_interfaces.human import HumanChatSession
-from model_interfaces.huggingface_interface import HFChatSession
-from model_interfaces.gemini_interface import GeminiProInterface
-from model_interfaces.memorybank_interface import MemoryBankInterface #Memory Bank
-from model_interfaces.memgpt_interface import MemGPTInterface
```
JosephDavidsonKSWH, line 24: Why the change to local imports?

ltm/agent_think.py
dcasbol: No. I just let them there for when we have the new evaluation tool. But now that you mention it, it seems dumb. I'll just save the file in my stash and remove it from here.

runner/run_benchmark.py
dcasbol: It all started with MemGPT, and when I couldn't get around its dependencies. Then I realized that the same could happen with MemoryBank or any of those 3rd party agents. With the local imports we could let the user to install only what they're actually using. What do you think?

ltm/agent.py
JosephDavidsonKSWH: So we need to sort out our nomenclature here. From the top the LTMAgentWrapper has `max_prompt_size` and `max_message_size`. 

- First off: What message does `max_message_size` refer to? The input or output?

In `LTMAgentWrapper` instantiates `LTMReflectionAgent` with `max_message_size -> max_completion_tokens`

- So now its the output message size?

In `LTMAgent` we have `max_input_tokens` which is  `0.9 * (max_prompt_size - completion_size)`, so that takes account of the output tokens of the model, OR the input size, but not both.

There is also a `max_message_size` at line 66 which is never used.

dcasbol: Right, I agree on the `max_message_size` issue. I think it should apply only to input messages, and we should have a much shorter length for completion. I applied the `0.9` factor in `max_input_tokens` because there were slight errors in early versions and I didn't want to deal with them, but we can hopefully get rid of it now. `max_message_size` was originally established to communicate the scheduler the agent's capability for taking inputs, so what about defining a `max_response_size` for the output?

runner/scheduler.py
JosephDavidsonKSWH: It does but, `making` -> `asking`

utils/llm.py
```diff
@@ -196,13 +193,13 @@ def log_llm_call(run_name: str, agent_name: str, debug_level: int, label: str =
...
     save_dir.mkdir(parents=True, exist_ok=True)
 
-    # Sanitize timestamp
-    sanitized_timestamp = datetime.now().strftime("%Y-%m-%d_%H_%M_%S_%f")
-
     # Write content of LLM call to file
     context = _llm_debug_params.pop("messages")
     response_text = _llm_debug_params.pop("response")
-    save_path = save_dir.joinpath(f"{sanitized_timestamp}{'-' + label if label is not None else ''}.txt")
+    save_name = datetime.now().strftime("%Y-%m-%d-%H-%M-%S-%f")
```
JosephDavidsonKSWH, line 199: Another unclear comment. `"%Y-%m-%d_%H-%M-%S-%f"` <- do this

JosephDavidsonKSWH (approved):

PR #148, opened 2024-09-06
Script to go back to a checkpoint and test an agent on that exact situation.

dcasbol:


dev_scripts/checkpoint_eval.py
```diff
@@ -0,0 +1,97 @@
...
+    agent.agent.hybrid_memory.semantic_memory.retrieve = patched_retrieve
+
+    # Same for recent messages
+    def patched_recent_messages(*args, **kwargs):
+        messages = patched_recent_messages.get_recent_messages(*args, **kwargs)
+        limit_timestamp = target_date.timestamp()
+        return [m for m in messages if m.timestamp < limit_timestamp]
+
+    patched_recent_messages.get_recent_messages = agent.agent.hybrid_memory.get_recent_messages
+    agent.agent.hybrid_memory.get_recent_messages = patched_recent_messages
```
JosephDavidsonKSWH, line 83: Nice solution!

JosephDavidsonKSWH:
There is also a bug in `RetrievalEvaluator.py:71`:

`self.project_root = DATA_DIR.join("retrieval_evaluator") ` ->  `self.project_root = DATA_DIR.joinpath("retrieval_evaluator")`

Can you also add that fix in there so that run_benchmark can run again?

dev_scripts/checkpoint_eval.py
dcasbol: 😁 I'm afraid we'll have to fix it at some point, since `retrieve` retrieves a limited amount of memories, and if these happen to be after the filtering date, the result might be quite off. It should work for the moment, though 🤞🏾

JosephDavidsonKSWH (approved):

